PR #7, opened 2016-09-08
Add 'textarea' core editor

geigerzaehler:
Tests will be provided at a later point.

### DIFF
--- a/core-field-editors/README.md
+++ b/core-field-editors/README.md
@@ -1,0 +1,36 @@
+Contentful Core Field Editors
+=============================
+
+This folder contains the source code for field editors builtin within the
+Contentful Web Application. Since these are developed using the
+[UI Extensions SDK][ui-extensions-sdk], this will allow you to understand how
+each component works, and create your own extensions based on existing
+Contentful components' source rather than starting from scratch.
+
+To do this, you should first clone the repository and install the field editor
+dependencies as follows:
+~~~
+git clone https://github.com/contentful/extensions.git contentful-extensions
+cd contentful-extensions/core-field-editors
+npm install
+npm install -g contentful-extension-cli
+~~~
+
+Then go to the field editor folder and create a bundle called `editor.dist.html`,
+which will be used as the extension's source:
+~~~
+cd src/textarea
+../../bin/bundle
+~~~
+
+Finally, create the extension from the `extension.json` definition and the HTML
+file.
+~~~
+export CONTENTFUL_MANAGEMENT_ACCESS_TOKEN=<MY_TOKEN>
+contentful-extension create --space-id <MY_SPACE_ID>
+~~~
+
+You can now use the component by following the same steps as with any other
+[UI extension](https://www.contentful.com/blog/2016/07/06/ui-extensions-sdk/).
+
+[ui-extensions-sdk]: https://github.com/contentful/ui-extensions-sdk
--- a/core-field-editors/bin/bundle
+++ b/core-field-editors/bin/bundle
@@ -1,0 +1,78 @@
+#!/usr/bin/env node
+
+/* eslint no-console: off */
+const P = require('path')
+const Browserify = require('browserify')
+const h = require('virtual-dom').h
+const vdomToString = require('vdom-to-html')
+const B = require('bluebird')
+const Yargs = require('yargs')
+
+const FS = B.promisifyAll(require('fs'))
+
+B.coroutine(main)().done()
+
+function* main () {
+  const args = parseCLI()
+  const entryPath = P.resolve(args.entryPath)
+  const stat = yield FS.statAsync(entryPath).catch(() => null)
+
+  if (!stat) {
+    console.error(`Cannot read file '${args.entryPath}'`)
+    process.exit(1)
+  }
+  const entryDir = stat.isDirectory() ? entryPath : P.dirname(entryPath)
+  const output = P.resolve(entryDir, 'editor.dist.html')
+  console.log(`Bundling field editor extension to '${P.relative('', output)}'`)
+  const bundle = yield makeBundle(entryPath)
+  const html = toHtml(editorTree(bundle))
+  yield FS.writeFileAsync(output, html, 'utf8')
+}
+
+function parseCLI () {
+  const argv = Yargs(process.argv.slice(2))
+    .strict()
+    .usage('bin/build [FILE]')
+    .argv
+
+  return {
+    entryPath: argv._[0] || '.',
+  }
+}
+
+function makeBundle (path) {
+  path = P.resolve(path)
+  return new B.Promise((resolve, reject) => {
+    Browserify({debug: false})
+    .transform('babelify')
+    .transform({global: true, sourcemap: false}, 'uglifyify')
+    .require(path, {expose: 'field-editor'})
+    .bundle((err, buf) => {
+      if (err) {
+        reject(err)
+      } else if (buf) {
+        resolve(buf.toString())
+      }
+    }).on('error', reject)
+  }).catch((err) => {
+    // Errors thrown by browserify are not instances of 'Error'.
+    throw new Error(err.toString())
+  })
+}
+
+function editorTree (bundle) {
+  return h('html', [
+    h('head', [
+      h('script', [bundle]),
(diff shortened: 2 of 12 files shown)


core-field-editors/Makefile
```diff
@@ -0,0 +1,7 @@
+export PATH := ./node_modules/.bin:${PATH}
+
+compile:
```
Khaledgarbaya: Hi, Couldn't this also be moved to the package.json as an npm script, plus npm will detect the right binary to use without specifying explicitly the PATH

core-field-editors/src/textarea/index.js
```diff
@@ -0,0 +1,40 @@
+import throttle from 'lodash/throttle'
+import createUpdater from '../_shared/input-updater'
+import mkBoot from '../_shared/boot'
+
+export const boot = mkBoot(create)
+
+export default function create (widgetApi, opts = {}) {
```
Khaledgarbaya: shouldn't be `extensionApi`, since we replaced the name `widget` by `extension`

core-field-editors/package.json
```diff
@@ -0,0 +1,43 @@
+{
```
Khaledgarbaya: one last comment from me, should we maybe add an npm [shrinkwrap](https://docs.npmjs.com/cli/shrinkwrap) so we make sure all users will have the exact same versions

geigerzaehler: Not a big fan of having this extra overhead of keeping the shrinkwrap in sync and clutter free.

core-field-editors/bin/bundle
```diff
@@ -0,0 +1,79 @@
...
+  }).catch((err) => {
+    // Errors thrown by browserify are not instances of 'Error'.
+    throw new Error(err.toString())
+  })
+}
+
+function editorTree (bundle) {
+  return h('html', [
+    h('head', [
+      h('script', {src: 'https://contentful.github.io/ui-extensions-sdk/cf-extension-api.js'}),
```
jelz: Could we use an npm version of  `ui-extensions-sdk`? Why do we prefer the hosted version over including it in the bundle? It seems a little bit fragile.

If bundle sizes are becoming an issue then maybe we should bump the limit?

If we want to use the hosted version then please extract `const BASE = 'https://contentful.github.io/ui-extensions-sdk/';`

core-field-editors/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "uglifyify": "^3.0.2",
+    "vdom-to-html": "^2.3.0",
+    "yargs": "^5.0.0"
+  },
+  "babel": {
+    "presets": [
+      "es2015"
+    ]
+  },
+  "dependencies": {
```
jelz: I'm not sure if we should split dependencies into two section in this kind of project. To do anything meaningful (bundle an editor) with a cloned repo you need to have all the deps.

core-field-editors/bin/bundle
```diff
@@ -0,0 +1,79 @@
...
+    .strict()
+    .usage('bin/build [FILE]')
+    .argv
+
+  return {
+    entryPath: argv._[0]
+  }
+}
+
+function mkBundle (path) {
```
jelz: Why `mk`? The same goes for `mkBoot` in `boot.js`

core-field-editors/src/textarea/index.js
```diff
@@ -0,0 +1,40 @@
...
+
+  field.onSchemaErrorsChanged(function (errors) {
+    if (errors && errors.length > 0) {
+      textarea.setAttribute('aria-invalid', true)
+    } else {
+      textarea.removeAttribute('aria-invalid')
+    }
+  })
+
+  const updateValue = throttle(function () {
```
jelz: Can we extract throttling setup to a helper function that could be used across editors (for consistent behavior).

We could also consider attaching events in such function (so it'd be `fn(el, onupdate)`). I think the code below applies to all textual editors.

core-field-editors/package.json
geigerzaehler: I think it is meaningful. If you are using the package as a library you are only going to use the transpiled files in `./lib`. And these only require the production dependencies. If you want to bundle an editor you are not going to use the NPM package but rather the repo with `npm install`.

core-field-editors/src/textarea/index.js
geigerzaehler: For now this is fine, I think. If we add more editors we can refactor.

core-field-editors/README.md
```diff
@@ -0,0 +1,32 @@
+Contentful Core Field Editors
+=============================
+
+This folder contains the source code to some of the Contentful Web App’s builtin
+field editors. The code targets the [Contentful UI Extension
+API][ui-extension-sdk]. This allows you to modify the source code, upload the
+editors as extensions to Contentful and use them in our App.
+
```
hlabas: Proposal: 
"This folder contains the source code for field editors builtin within the Contentful Web Application.
Since these are developed using the UI Extensions SDK, this will allow you to understand how each component works, and create your own extensions based on existing Contentful components' source rather than starting from scratch."

core-field-editors/README.md
```diff
@@ -0,0 +1,32 @@
...
+To do this clone the repository and install the field editor dependencies.
```
hlabas: Proposal:
"To do this, you should first clone the repository  and install the field editor dependencies as follows:"

core-field-editors/README.md
```diff
@@ -0,0 +1,32 @@
...
+~~~
+git clone https://github.com/contentful/extensions.git contentful-extensions
+cd contentful-extensions/core-field-editors
+npm install
+npm install -g contentful-extension-cli
+~~~
+
+Move to the field editor folder and bundle the source code.
```
hlabas: Proposal:
"Then go to the field editor folder and create a bundle called `editor.dist.html`, which will be used as the extension's source:"

core-field-editors/README.md
```diff
@@ -0,0 +1,32 @@
...
+~~~
+cd src/textarea
+../../bin/bundle
+~~~
+This will create the `editor.dist.html` file including the bundled and inlined
+code.
+
```
hlabas: I'd remove these two lines if using the above proposal.

core-field-editors/README.md
```diff
@@ -0,0 +1,32 @@
...
+Finally, create the extension from the `extension.json` definition and the HTML
+file.
+~~~
+export CONTENTFUL_MANAGEMENT_ACCESS_TOKEN=<MY_TOKEN>
+contentful-extension create --space-id <MY_SPACE_ID>
+~~~
+
```
hlabas: I'd add: "You can now use the component by following the same steps as with any other [UI extension](https://www.contentful.com/blog/2016/07/06/ui-extensions-sdk/)"

core-field-editors/src/_shared/boot.js
```diff
@@ -0,0 +1,10 @@
+import {init} from 'contentful-ui-extensions-sdk'
+
+export default function makeBooter (create) {
+  return function boot () {
+    init((api) => {
+      api.window.startAutoResizer()
```
hlabas: Doing this by default doesn't feel super right to me since auto-sizing doesn't work well with dropdowns or absolutely positioned pop-ins. Explicit calls make that somehow clearer IMO.

hlabas:
Do we go for `core-field-editors/src/[name]` or `core-field-editors/[name]/src/`?
I'm asking as one main interest would be to fork a specific component and not necessarily the entire repo. 
This might require a discussion around versioning each component or the entire set of components at once.
I'd ideally go for individual components since they evolve at different paces.

geigerzaehler:
> Do we go for core-field-editors/src/[name] or core-field-editors/[name]/src/?

I’d stick with the current layout. In particular because I expect the components to share some functionality. It also makes managing the (yet unpublished) NPM package easier. We can revisit this later.

core-field-editors/README.md
geigerzaehler: :heavy_check_mark:

hlabas:
LGTM

jelz:
LGTM

PR #18, opened 2017-03-02
Makes json-form-editor load before cfApi

lwakefield:
Loading `json-form-editor` after `api` has been initialized means you can run the example without having a connection to contentful api. Of course, this means no data will be propagated locally.

### DIFF
--- a/samples/json-form-editor/src/json-form-editor.js
+++ b/samples/json-form-editor/src/json-form-editor.js
@@ -7,37 +7,36 @@
 
 var cfExt = window.contentfulExtension || window.contentfulWidget;
 
-cfExt.init(initContentfulJsonFormEditor);
+var cfApi = null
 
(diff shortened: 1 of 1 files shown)


DanweDE:
Hi lwakefield. Thank you for your contribution.

I am curious, what's the use-case for running the example without having a connection to the contentful API?

lwakefield:
Hey.

Unfortunately, it has been quite a while since I worked with Contentful Extensions. From memory however, I wanted to make some changes to json-form-editor for a custom case. In order to develop, it needed to be connected to Contentful, which as you can imagine, the edit, redploy, test cycle was painful.

lwakefield:
I'm happy for you to close this if you think that makes sense.

jelz:
Closed, we'll revamp this extension soon.

PR #26, opened 2017-12-18
Fix Extentions API link in README.md

christineywang:
When I installed the extension, I was getting this error:
```Cannot read the file defined as the "srcdoc" property (/Users/christinewang/Documents/Coding/extensions-master/samples/youtube-id/dist/index.html).```

There appears to be an error in the file path and it worked when I removed the dist path.


christineywang:
Also adding a fix to a broken api doc link.

christineywang:
Per your request, I've updated my commit history.

PR #29, opened 2018-01-19
Add external API sample

thehighlander:
Demonstrates exposing external API data within the Contentful UI.

### DIFF
--- a/samples/external-api/README.md
+++ b/samples/external-api/README.md
@@ -1,0 +1,2 @@
+# Contentful UI Extension 
+Demonstrates exposing external API data within the contentful UI.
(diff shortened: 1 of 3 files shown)


floelhoeffel:
Thanks for the PR @thehighlander! Could you please add a screenshot and a short description to the readme? Doesn't have to be long, maybe something like

- fetches dummy person data from this endpoint
- writes the data into a dropdown
- on select, the id of the selected dummy user is stored as the field value in contentful

thehighlander:
Gladly @floelhoeffel! I wanted exactly this sample so I thought I'd just provide it for the next guy.

floelhoeffel:
Thanks @thehighlander for the contribution - we will change field type to support short text fields and add a screenshot

PR #31, opened 2018-01-25
Improved external API example

floelhoeffel:
- added screenshot
- using DOM methods 
- autosize for iframe


samples/external-api/app.html
```diff
@@ -1,43 +1,56 @@
 <!doctype html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8"/>
+
+<head>
+    <meta charset="UTF-8" />
```
jelz: No need for self-closing tag `/>`

samples/external-api/app.html
```diff
@@ -1,43 +1,56 @@
...
-        });
-
-        $("#user-id").on('input', function() {
-            dd.setValue(this.value);
-        });
-      })
+        window.contentfulExtension.init(function (api) {
+            api.window.startAutoResizer();
+            var value = api.field.getValue()
+            var dd = api.field;
```
jelz: `dd`? Either name it `field` or always use `api.field` which is also quite ok for me.

samples/external-api/app.html
```diff
@@ -1,43 +1,56 @@
...
+            var serviceUri = "https://jsonplaceholder.typicode.com/users";
+
+            $.ajax({
+                url: serviceUri,
+                type: "GET",
+                crossDomain: true,
+                dataType: "json",
+                success: function (response) {
+                    $.each(response, function (index, value) {
```
jelz: `response.forEach(function () { ... })` would be better

samples/external-api/app.html
```diff
@@ -1,43 +1,56 @@
...
+
+                        var option = document.createElement("option")
+                        option.setAttribute("value", value.id)
```
jelz: Use or don't use `;` but be consistent

samples/external-api/app.html
```diff
@@ -1,43 +1,56 @@
...
+                        option.innerText = value.name
+
+                        $("#user-id").append(option);
+                    });
+                    $("#user-id").val(dd.getValue());
+                },
+                error: function (xhr, status) {
+                    console.log("error");
```
jelz: `error fetching data`. You could also include status code and response.

samples/external-api/app.html
axe312ger: 👍

floelhoeffel:
Implemented your suggestions - thanks @jelz

PR #33, opened 2018-01-26
Updated the diff extension.

floelhoeffel:
-Migrated example to new CLI
-fixed a small issue with empty published values
-simplified readme so extension management instructions are not duplicated

### DIFF
--- a/samples/diff/README.md
+++ b/samples/diff/README.md
@@ -6,7 +6,7 @@ The diff editor extension shows the diff between the draft value and the publish
 
 ## Installation and usage
 
-Check you have the [requirements](../README.md#requirements) needed to use our extensions.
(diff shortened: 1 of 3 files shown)


samples/diff/README.md
```diff
@@ -41,22 +16,14 @@ Start a local server, changing the port if needed:
 python -m SimpleHTTPServer 3030
 ```
 
-Tell Contentful to render the widget from your local machine:
+Don't use the `extension.json` descriptor file but instead use 3rd party hosting from `localhost`
 
 ```bash
-npm run dev
+contentful extension update --force --src 'http://localhost:3030/index.html' --id diff --name diff -field-types Symbol -field-types Text
```
axe312ger: `--field-types`

floelhoeffel: 👍

PR #53, opened 2018-05-06
Update samples to use sdk v3

floelhoeffel:


floelhoeffel:
Migrated all samples except chessboard and json-form-editor.

PR #55, opened 2018-05-09
Parameterize shopify UI extension

stefanjudis:


samples/shopify/env.example
```diff
@@ -1,4 +1,2 @@
 export SPACE_ID=000
-export CONTENTFUL_MANAGEMENT_ACCESS_TOKEN=000
-export shopifyUrl='https://demo.myshopify.com/api/graphql`
-export shopifyToken=000
\ No newline at end of file
+export CONTENTFUL_MANAGEMENT_ACCESS_TOKEN=000
```
floelhoeffel: why does this need a CMA token?

PR #81, opened 2018-12-06
Migrate external-api example to React

m10l:
This PR migrates the external-api example to React so that we're able to use React components from Forma 36 to build an extension.

screenshot 2018-12-06 at 13 26 12

### DIFF
--- a/samples/external-api/index.html
+++ b/samples/external-api/index.html
@@ -5,14 +5,14 @@
     <title>External API UI Extension Sample</title>
     <link
       rel="stylesheet"
-      href="//unpkg.com/@contentful/forma-36-react-components@1.0.1/dist/styles.css"
+      href="//unpkg.com/@contentful/forma-36-react-components@1.0.3/dist/styles.css"
     />
     <script src="//unpkg.com/whatwg-fetch@3.0.0/dist/fetch.umd.js"></script>
     <script src="//unpkg.com/contentful-ui-extensions-sdk@3"></script>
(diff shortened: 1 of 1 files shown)


samples/external-api/index.html
```diff
@@ -0,0 +1,93 @@
...
+            this.state.hasLoaded &&
+              h(
+                SelectField,
+                {
+                  name: "names",
+                  id: "names",
+                  labelText: "Select name",
+                  helpText: "Get dummy user data from an external API",
+                  onChange: this.onChange,
+                  value: this.state.value
```
jelz: How about having a default and an option for it? "Pick dummy user..."

samples/external-api/index.html
```diff
@@ -0,0 +1,93 @@
...
+                },
+                this.state.items.map(item =>
+                  h(
+                    Option,
+                    { key: item.id, value: item.id.toString() },
```
jelz: why `toString`? It feels kind of unsafe to externally provided data. I tend to use:

```js
`${item.id}`
```

samples/external-api/index.html
```diff
@@ -0,0 +1,93 @@
+<!DOCTYPE html>
```
jelz: I'd mention in the README that this extension requires a modern browser. I think up until this PR we were trying to write code so it works in IE11.

suevalov: @jelz I updated PR and made it ES5 only

Issue #63, opened 2018-07-05
Slug Gen docs don't explain the extention


emilbruckner:
I added the slug generator to a project a couple of months ago and had no problems, added it to another one today, and this happened:

![image](https://user-images.githubusercontent.com/9105589/42343269-d8481b98-8098-11e8-83ea-7c3690b0121a.png)

The slug won't generate and all three icons show up in it.
I read this error: `Cannot read property 'onValueChanged' of undefined`

The problem is in this line of the plugin:
```javascript
var titleField = api.entry.fields.title
```

The generator assumes that the title is the field with field-id `title`. Wasn't the case for this project. 

### How to solve:
Either
1) Get the title from the entry title, not the field `title`
2) Update the README

suevalov:
Solved by #83

PR #83, opened 2019-01-09
Slug Gen - Use entry title for title field

Cruzercru:
Rather than using a set name field (title) to generate slugs from, they will be generated from the entry title field. Closes issue: https://github.com/contentful/extensions/issues/63#issue-338697167


suevalov:
Thanks @Cruzercru!

PR #93, opened 2019-03-09
A place for Marketplace-ready extensions

realityking:
This PR adds a new folder to the repo called `marketplace`. It's intended to hold those extensions that are hosted in the Marketplace and are not examples but ready for production.

Still some refinements need to be done but everything works :)

Missing things:

- [x] Local dev mode comparable with `create-contentful-extension`
- [ ] Sensitive Language extension needs some options before it can go into the Marketplace


floelhoeffel:
Oh, that is a good idea. I wanted to separate proper extensions from
samples anyway.

scripts/build.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+
+const Bundler = require('parcel-bundler');
```
suevalov: I think we should not reinvent the way we bundle extensions. 

I think that this script should just simply go through the list of extensions that are intended to be hosted on a separate domain. All `build/index.html` files should be committed as they are now. The script would read `version` in package.json of each extension and would push the content of `build/index.html`, tagged by current version, to Netlify or whatever.

marketplace/sensitive-language/src/index.css
```diff
@@ -0,0 +1,70 @@
+@font-face {
+  font-family: "Avenir Next W01";
+  src: url("https://static.contentful.com/app/2cd55546-ec00-4af9-aeca-4a3cd186da53.woff2")
```
jelz: don't use avenir in open-source

package.json
```diff
@@ -0,0 +1,54 @@
...
+    "coverage": "jest --coverage marketplace/**/test/*.test.js"
+  },
+  "engines": {
+    "node": ">=10.15.0"
+  },
+  "dependencies": {
+    "@contentful/forma-36-fcss": "0.0.13",
+    "@contentful/forma-36-react-components": "^2.9.8",
+    "@contentful/rich-text-plain-text-renderer": "^13.1.0",
+    "alex": "^7.1.0",
```
jelz: did I miss something or we defined only one package.json file for the whole repo?

jelz:
What I would propose:

1. `marketplace` holds independent extensions. Each subdirectory of `marketplace` has its own dependencies and tests; you should be able to move one subdirectory (`marketplace/sensitive-language` for example) and still use it (as in: install deps, run tests, build and dev-deploy)
2. Extensions are created with create-contentful-extension
3. Built artifacts are not committed
4. CI iterates through `marketplace/*` and does `npm i && npm test && npm build` for each
5. We collect built artifacts in CI and deploy

scripts/deploy.js
```diff
@@ -0,0 +1,80 @@
...
+const logSymbols = require('log-symbols');
+
+const {dirs} = require('./utils.js');
+
+const BUILD_DIR = path.join(__dirname, '..', 'dist');
+const client = new NetlifyAPI(process.env.NETLIFY_TOKEN);
+
+function createOrUpdate (extensionName) {
+  const siteName = `ctf-exts-${extensionName}`;
+  const customDomain = `${extensionName}.contentfulexts.com`;
```
jelz: why do we need a site per extension again?

realityking: To allow for breaking changes. This way whenever we have to/want to make a breaking change we update the `majorVersion` property in `extension.json` which will automatically create a new subdomain. The old subdomain won't be touched again by this repository and will work forever. This way we don't impact existing installations.

A good example of why braking changes are sometimes necessary are sidebar extensions. If we had any extensions using the old sidebar model in the Marketplace, now would be the time to update the `majorVersion` to use the new API.
 
In theory, Netlify does allow for incremental updates and maintaining previous files. But it's far outside their happy path and I'd rather not rely on it.

marketplace/sensitive-language/src/index.css
realityking: Will remove it 👍 

@m10l I guess this answers our discussion

package.json
realityking: Yeah I build it with a single package.json. It makes it easier to share things, especially for testing and linting. Definitely possible to break this up into individual packages, just a bit more involved in terms of set up.

I guess it comes down to how we want to maintain these extensions. If there's a new version of say Forma 36 released, do we wanna update it all in one go or do we see each extension as a fully separate project maintained independently?

scripts/build.js
realityking: I strongly disagree with storing build artefacts in git. It's a massive anti-pattern we currently do for Marketplace extensions because we don't have a better alternative. Committed build artefacts have a tendency to get out of sync with their src and you end up with weird commits when the build result changes because of the host environment of the committer.

realityking:
@jelz 

> marketplace holds independent extensions. Each subdirectory of marketplace has its own dependencies and tests; you should be able to move one subdirectory (marketplace/sensitive-language for example) and still use it (as in: install deps, run tests, build and dev-deploy)

That's definitely possible. See my above comment why I didn't go for this initially but I'm not married to my approach.

> Extensions are created with create-contentful-extension

`create-contentful-extension` doesn't currently support `src` extensions, does it? I'd also argue that for src extensions we might want slightly different behaviour. E.g. I'd probably rather not inline my 800kB JS for `sensitive-language`, I'd rather see it cached based on its hash.

Do we wanna add these options to `create-contentful-extension`?

> Built artifacts are not committed

👍 

> CI iterates through marketplace/* and does npm i && npm test && npm build for each
We collect built artifacts in CI and deploy

As above, it's a bit more complex but no principle opposition if you an Alex think this a better approach.

scripts/build.js
suevalov: Building extension using `create-contentful-cli` on a fly would also work. I just don't want to have another way to build things.

package.json
suevalov: I think each extension should be self contained, so we can change dependencies per extension.

suevalov:
@realityking "Do we wanna add these options to create-contentful-extension?" Yes, it's planned as an optional build behavior: `contentful-extension-scripts build --no-inline`

realityking:
@suevalov Awesome!

Any idea how to configure it to use `src` instead of `srcdoc`? Or should I rewrite it like I currently do?

suevalov:
@realityking Rewriting `extension.json` before publishing seems reasonable to me.

scripts/deploy.js
jelz: Alternative that doesn't produce all those Netlify sites would be to just write versioned directories to an S3 bucket and put a Cloudfront distribution in front of it.

realityking:
@jelz @suevalov I believe I've addressed the comments regarding the build process. Please take another look 🙂

scripts/deploy.js
realityking: That would work. Though I actually quite like the fully independent sites. I've checked with Jani before building this, and he's fine with ~30 extra sites in the main account. If it becomes a problem, we can always move them to their own Netlify account.

Bonus advantage, we can use Netlify functions for some extensions. I got my eyes on the Marketo one for example.

realityking:
Quick note if you're testing this, `eslint` _only_ works before running `npm run bootstrap`. Afterwards it gets confused by the `node_modules` folders 😔

suevalov:
@realityking `.eslintignore` to the rescue: https://eslint.org/docs/user-guide/configuring#eslintignore

What do you think about adding `lint-staged` and `prettier`? I can jump in and help.

realityking:
@suevalov Thank you!

> .eslintignore to the rescue: https://eslint.org/docs/user-guide/configuring#eslintignore

That set me on the right path. I forgot that eslint doesn't use `.gitignore` by default

> What do you think about adding lint-staged and prettier? I can jump in and help.

I'll add `prettier` now but I think `lint-staged` can wait. Can you take another look over the build/deploy script? I'd love to merge this soon so I can pivot to making the necessary changes to the `doc-app`.

realityking:
Prettier added 🎉

.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
suevalov: Can you, please, make it consistent with `.prettierrc` in the web app?

marketplace/google-analytics/src/app.js
```diff
@@ -0,0 +1,135 @@
...
+import {formatDate} from './utils.js';
+
+const DAY_IN_MS = 1000 * 60 * 60 * 24;
+const TODAY = new Date();
+const TIMELINE_DIMENSIONS = [
+  {label: 'Day', value: 'date'},
+  {label: 'Week', value: 'week'},
+  {label: 'Month', value: 'month'}
+];
+const CLIENT_ID = '318721834234-s3td95ohvub1bkksn3aicimnltvmtts8.apps.googleusercontent.com';
```
suevalov: Shouldn't it be a parameter?

marketplace/google-analytics/src/app.js
```diff
@@ -0,0 +1,135 @@
...
+      isAuthorized: false,
+      hasSlug,
+      range: {
+        start: new Date(TODAY - DAY_IN_MS * 14),
+        end: TODAY
+      },
+      dimension: 'date',
+      pagePath
+    };
+    auth.on('signIn', () => this.setState({isAuthorized: true}));
```
suevalov: Looks like a candidate for `componentDidMount`

marketplace/google-analytics/src/app.js
```diff
@@ -0,0 +1,135 @@
...
+                  onChange={this.handleDimensionChange.bind(this)}
+                  checked={isActive}
+                />
+              </label>
+            );
+          })}
+        </div>
+        <Timeline pagePath={pagePath} range={range} dimension={dimension} viewId={parameters.viewId} />
+        <div className="info">
+          <svg fill="#2D2F31" width="100pt" height="100pt" viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg">
```
suevalov: Would be nice to extract this icon to a separate component

marketplace/sensitive-language/src/no-issues.js
```diff
@@ -0,0 +1,19 @@
...
+
+import React from 'react';
+
+import {Icon, Typography, Subheading, Paragraph} from '@contentful/forma-36-react-components';
+
+export function NoIssues() {
+  return (
+    <Typography>
+      <Subheading style={{display: 'flex', alignItems: 'center'}}>
+        <Icon icon="CheckCircle" color="positive" extraClassNames="f36-margin-right--xs" />
```
suevalov: I see too many to ways to apply styles here: `Typography` component, inline styles and forma utility classes. Would be better just to use CSS in this case.

marketplace/sensitive-language/src/unsupported-language.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {Icon, Typography, Subheading, Paragraph} from '@contentful/forma-36-react-components';
+
+export function UnsupportedLanguage({localeCode, localeName}) {
+  return (
+    <Typography>
+      <Subheading style={{display: 'flex', alignItems: 'center'}}>
```
suevalov: The same as above

marketplace/google-analytics/src/app.js
realityking: Why? That would require the user to create their own oAuth app. This way it's already done which saves _a lot_ of hassle setting up the app.

marketplace/google-analytics/src/app.js
realityking: Went ahead and used the Forma 36 component instead :)

marketplace/sensitive-language/src/no-issues.js
realityking: @m10l Since you wrote this part, WDYT?

realityking:
@suevalov Removed the sensitive language extension as per our Slack conversation earlier. Will make a new pull request after this is merged so we can focus on one extension at a time.

marketplace/google-analytics/extension.json
```diff
@@ -0,0 +1,32 @@
+{
+  "id": "google-analytics",
+  "name": "google-analytics",
+  "majorVersion": 1,
```
suevalov: Just noticed it. I wouldn't add new properties that are incompatible with API in the extension.json. Can you use `package.json` version for this purpose?

realityking:
@suevalov Done! Please take another - and hopefully last - look :)

PR #95, opened 2019-03-13
Image Uploader

azer:


samples/image-uploader/src/components/ProgressView/index.js
```diff
@@ -0,0 +1,23 @@
+import React from "react"
+import "./progress-view.css"
+
+export default function UploadView(props) {
```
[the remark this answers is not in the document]
azer: Skipped this as that method now takes more props.

samples/image-uploader/src/index.js
```diff
@@ -0,0 +1,437 @@
...
+    // Handler for external field value changes (e.g. when multiple authors are working on the same entry).
+    this.detachExternalChangeHandler = this.props.sdk.field.onValueChanged(
+      this.onExternalChange
+    )
+
+    if (this.state.value) {
+      this.props.sdk.space
+        .getAsset(this.state.value.sys.id)
+        .then(asset => this.setState({ asset }))
+        .catch(this.onError)
```
hwartig: How do you feel about improving the error reporting here. As `value.sys.id` might reference a deleted asset `getAsset` might fail not only because of network errors. In case of a deleted asset we could either

a) clear the current value in the field and show the extension as if the field was empty in the first place (this might lead to problems if there is a validation on the field)
b) show the user an appropriate error message that indicates that he needs to remove the current value and upload or reference another asset.

In any case we should allow the user to make changes to the value of the field. However the current experience just shows an endless spinner and a very generic "request failed error"

Screenshot 2019-03-20 at 15 52 27

samples/image-uploader/src/components/UploadView/upload-view.css
```diff
@@ -0,0 +1,49 @@
+.upload-view {
+  border: 2px dashed #ccc;
+  text-align: center;
+  background: #fff;
+  color: #888;
+}
+
+.upload-view main {
```
suevalov: dead code

samples/image-uploader/src/utils.js
```diff
@@ -0,0 +1,89 @@
...
+    prefix: parts[0],
+    data: parts[1],
+    type: src.split(/:|;/)[1]
+  }
+}
+
+// getMimeTypeByPath(path: string): [mimetype, isImage]
+export function getMimeTypeByPath(uri) {
+  // Unsplash doesn't have file extensions in their urls, so
+  if (isUnsplashImageUrl(uri)) {
```
suevalov: isn't it a too specific check? there are thousands of websites

samples/image-uploader/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "@babel/plugin-proposal-class-properties": "7.3.4",
+    "@babel/plugin-transform-runtime": "7.3.4",
+    "@babel/preset-env": "7.3.4",
+    "@babel/preset-react": "7.0.0",
+    "@contentful/contentful-extension-scripts": "0.4.2",
+    "contentful-cli": "0.20.0"
+  },
+  "dependencies": {
+    "@contentful/forma-36-fcss": "0.0.13",
+    "@contentful/forma-36-react-components": "2.9.11",
```
suevalov: I'd bump it to `2.10.1` with the latest fixes and better font fallbacks

samples/image-uploader/src/utils.js
hwartig: How about we fire a `HEAD` request and look at the `Content-Type` header if we cannot infer the mime type from the url?

samples/image-uploader/src/index.js
```diff
@@ -0,0 +1,437 @@
...
+    if (files.length > 1) {
+      return this.onError(new Error("Please drop only one image at a time"))
+    }
+
+    const file = files[0]
+    this.setState({ file })
+    this.setUploadProgress(0)
+
+    const [contentType, isImage] = getMimeTypeByPath(file.name)
+    if (!isImage) {
```
hwartig: I feel we have to support all asset types properly. Assets can change their type later. So if someone by accident replaces the image file within an asset with a pdf doc the image uploader right now will break pretty badly. Another example when this can happen is if there are already existing entries with existing references to non image assets before the image uploader was installed. I suggest we fully support all assets. However if we opt not to support all types of assets than we should at least not break the image uploader if it is confronted with non image assets.

samples/image-uploader/src/index.js
azer: @hwartig good point, fixed

samples/image-uploader/src/utils.js
azer: Removed mime type sniffing in favor of relying on `DataTransfer` content

samples/image-uploader/src/index.js
azer: Fixed

PR #96, opened 2019-03-20
Bynder Dialog

suevalov:


samples/bynder-dialog/src/index.js
```diff
@@ -0,0 +1,97 @@
...
+import { init, locations } from 'contentful-ui-extensions-sdk';
+import '@contentful/forma-36-react-components/dist/styles.css';
+import './index.css';
+import extension from '../extension.json';
+import BynderDialog from './BynderDialog';
+import Thumbnail from './Thumbnail';
+
+function reducer(state, action) {
+  switch (action.type) {
+    case 'add':
```
lehnerm: `add` is never dispatched and can easily be replaced with `add-all` anyway;

I'd make `state` an object, with the array being an items property to have an easier time adding additional state elements at a later stage.

In addition, the naming of the actions is very short and does not completely explain what they do; How about `add-all-assets` and `remove-asset` or so?

samples/bynder-dialog/src/index.js
```diff
@@ -0,0 +1,97 @@
...
+      sdk.window.stopAutoResizer();
+    };
+  }, []);
+
+  useEffect(() => {
+    sdk.field.setValue(assets);
+  }, [assets]);
+
+  const onButtonClick = async () => {
+    const assets = await sdk.dialogs
```
lehnerm: Can we somehow get rid of the scrollbar for the dialog? Seems to be caused by improper bynder div sizes.

samples/bynder-dialog/src/index.js
```diff
@@ -0,0 +1,97 @@
...
+    return () => {
+      sdk.window.stopAutoResizer();
+    };
+  }, []);
+
+  useEffect(() => {
+    sdk.field.setValue(assets);
+  }, [assets]);
+
+  const onButtonClick = async () => {
```
lehnerm: Can we rename this to `onBynderDialogOpen`?

samples/bynder-dialog/src/BynderDialog.js
```diff
@@ -0,0 +1,51 @@
...
+
+  componentDidMount() {
+    this.loadBynderScript();
+    this.props.sdk.window.updateHeight();
+    document.addEventListener('BynderAddMedia', e => {
+      const assets = e.detail.map(asset => ({
+        id: asset.id,
+        src: asset.thumbnails['webimage'],
+      }));
+      this.props.sdk.close(assets || []);
```
lehnerm: Default value is redundant here, if we want to be safe I'd suggest to check whether the event actually included an array:
```
      const eventDetail = Array.isArray(e.detail) ? e.detail : [];
      const assets = eventDetail.map(asset => ({
        id: asset.id,
        src: asset.thumbnails['webimage'],
      }));
      this.props.sdk.close(assets);
```

samples/bynder-dialog/src/index.js
```diff
@@ -0,0 +1,97 @@
...
+    ReactDOM.render(
+      <BynderDialog sdk={sdk} />,
+      document.getElementById('root')
+    );
+  } else {
+    ReactDOM.render(<Field sdk={sdk} />, document.getElementById('root'));
+  }
+});
+
+if (module.hot) {
```
lehnerm: Is this something that should go into the extension template by default?

samples/bynder-dialog/build/index.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+  <link rel="stylesheet" href="/src.e31bb0bc.css"></head>
+  <body>
+    <div id="root"></div>
+    <script type="text/javascript" src="/src.e31bb0bc.js"></script>
```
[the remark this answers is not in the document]
suevalov: Will be done soon https://github.com/contentful/extensions/pull/93

.eslintrc
```diff
@@ -1,14 +1,12 @@
 {
-  "extends": [
-    "eslint:recommended",
-    "plugin:promise/recommended",
-    "prettier"
-  ],
-  "plugins": [
-    "promise"
-  ],
+  "parser": "babel-eslint",
```
realityking: Can we rewrite this file to be an extension of `marketplace/.eslintrc`? Otherwise new rules we add won't be applied to this extension.

Even better, IMHO, would be if we could either have the code changed to conform to the rules or update `marketplace/.eslintrc` to have consistent code quality across the extensions in the `marketplace` directory.

marketplace/bynder-dialog/.gitignore
```diff
@@ -0,0 +1,19 @@
+# Logs
```
realityking: This file should be necessary since we have one in the root directory.

marketplace/bynder-dialog/src/BynderDialog.js
```diff
@@ -0,0 +1,54 @@
...
+      const results = Array.isArray(e.detail) ? e.detail : [];
+      const assets = results.map(asset => ({
+        id: asset.id,
+        src: asset.thumbnails['webimage']
+      }));
+      this.props.sdk.close(assets);
+    });
+  }
+
+  loadBynderScript = () => {
```
realityking: This doesn't need to be an arrow function. (doesn't use `this`)

Also why do this `componentDidMount` instead of adding the JS file directly to the HTML?

marketplace/bynder-dialog/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+    default:
+      return state;
+  }
+}
+
+function Field({ sdk }) {
+  const [assets, dispatch] = useReducer(reducer, sdk.field.getValue() || []);
+
+  useEffect(() => {
+    sdk.window.startAutoResizer();
```
realityking: I noticed this already in `create-contentful-extension`. Is there any advantage to doing it in the React lifecycle instead of doing it in init (before rendering)?

marketplace/bynder-dialog/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+    case 'add-all-assets':
+      return [...state, ...action.payload];
+    case 'remove-asset':
+      return state.filter(item => item.id !== action.payload);
+    default:
+      return state;
+  }
+}
+
+function Field({ sdk }) {
```
realityking: I suggest moving the component to its own field to leave the index file only for the routing/mounting code.

Also makes adding tests a lot easier.

marketplace/bynder-dialog/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+
+init(sdk => {
+  if (sdk.location.is(locations.LOCATION_DIALOG)) {
+    ReactDOM.render(<BynderDialog sdk={sdk} />, document.getElementById('root'));
+  } else {
+    ReactDOM.render(<Field sdk={sdk} />, document.getElementById('root'));
+  }
+});
+
+// if (module.hot) {
```
realityking: Why is this commented out?

marketplace/bynder-dialog/src/index.js
suevalov: In this particular case I don't want it to be called in Dialog extension and I use `updateHeight` there.

marketplace/bynder-dialog/src/index.js
suevalov: Removed

marketplace/bynder-dialog/src/index.js
suevalov: done

.eslintrc
```diff
@@ -1,14 +1,11 @@
...
-    "prettier"
-  ],
-  "plugins": [
-    "promise"
-  ],
+  "extends": ["eslint:recommended", "plugin:promise/recommended", "prettier"],
+  "plugins": ["promise"],
   "parserOptions": {
-    "ecmaVersion": 2019
+    "ecmaVersion": 8
```
realityking: Isn't `8` equivalent to `2017`? Why the regression?

.eslintrc
```diff
@@ -1,14 +1,11 @@
...
   },
-  "root": true
+  "root": true,
+  "rules": {
+    "promise/no-nesting": "off"
```
realityking: What's the reason to disable this rule? I don't see it being violated in the bynder extension. (I might've missed it)

package.json
```diff
@@ -33,5 +36,21 @@
   "engines": {
     "node": ">=10.15.0"
   },
+  "lint-staged": {
+    "*.js": [
+      "prettier --write --ignore-path .gitignore",
```
realityking: I just noticed this is not scoped to the `marketplace` directory. IMHO we should keep it just to that directory for now since the other folders are not following prettier formatting right now.
(same goes for the line below)

.eslintrc
suevalov: `extensions/scripts/build.js` violates this

.eslintrc
suevalov: reverted the change

package.json
suevalov: fixed

PR #98, opened 2019-03-25
Add help if extension is configured in wrong location

lehnerm:
Check whether location is entry-level sidebar; if not display a short help text


samples/ai-image-tagging/src/index.js
```diff
@@ -85,14 +85,15 @@ class App extends React.Component {
...
+    const { contentType, field, location, parameters: { instance: { imageFieldId, tagFieldId } } } = this.props.sdk;
     const { loadingTags, overwriteTags } = (this.state || {});
 
     const hasImageField = isCompatibleImageField(getField(contentType, imageFieldId));
     const hasTagField = isCompatibleTagField(getField(contentType, tagFieldId));
+    const isInCorrectLocation = location.is(locations.LOCATION_ENTRY_SIDEBAR);
 
     return (<div className='f36-color--text-light'>{
-      hasImageField && hasTagField ?
+      isInCorrectLocation && hasImageField && hasTagField ?
```
suevalov: This case logic seems to be complicated to me. I think it would be better to extract `isInCorrectLocation` from `ImageTaggingHelp` component and do something like this:

```
if (!location.is(locations.LOCATION_ENTRY_SIDEBAR)) {
   return <Note>The ai-image-tagging extension is an entry-level sidebar extension. You are currently using it on a field level.</Note>
}
return (
   <div className='f36-color--text-light'>
     {hasImageField && hasTagField ? <Component/> : < ImageTaggingHelp />}
   </div>
)


```

lehnerm: Yeah makes sense, like using note for this as well.

lehnerm: Extracted everything into their own components now, should be a lot easier to read

samples/ai-image-tagging/src/components/ImageTagging.js
```diff
@@ -0,0 +1,107 @@
...
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      overwriteTags: true,
+    }
+  }
+
+  componentDidMount() {
+    this.props.sdk.window.startAutoResizer();
```
suevalov: As far as I see `this.props.sdk.window.startAutoResizer();` is already called once in `src/index.js`.

PR #104, opened 2019-03-27
Add Approval Workflow extension

jelz:


samples/basic-approval-workflow/src/app.js
```diff
@@ -0,0 +1,249 @@
...
+    if (typeof this.off === 'function') {
+      this.off();
+    }
+  }
+
+  render = () => {
+    const { log, users } = this.props;
+    const { currentVersion, busy } = this.state;
+
+    const last = log[0];
```
zcei: As we're destructuring everywhere above:

```suggestion
    const [last] = log;
```

samples/basic-approval-workflow/src/app.js
```diff
@@ -0,0 +1,249 @@
...
+          <Button
+            buttonType="negative"
+            onClick={this.cancelReviewRequest}
+            isFullWidth
+          >
+            Cancel review request
+          </Button>
+        </React.Fragment>
+      );
+    } if (isReviewer) {
```
zcei: This should be either an `else if` or (even better) just on a newline 😆

samples/basic-approval-workflow/src/app.js
```diff
@@ -0,0 +1,249 @@
...
+
+    const item = {
+      ...data,
+      eventType,
+      id: id(),
+      userId: sdk.user.sys.id,
+      version: currentVersion,
+    };
+
+    // Send a message for real-time sync and persistance
```
zcei:
```suggestion
    // Send a message for real-time sync and persistence
```

samples/basic-approval-workflow/src/pubnub-client.js
```diff
@@ -0,0 +1,85 @@
...
+
+  function addToLog({ entry, timetoken }) {
+    const validEntry = entry && typeof entry.id === 'string';
+    const validTimetoken = typeof timetoken === 'string';
+    const newEntry = !state.log.find(cur => cur.id === entry.id);
+
+
+    if (validEntry && validTimetoken && newEntry) {
+      // Timetoken is a 17-digit precision unix time (UTC)
+      const t = parseInt(timetoken, 10) / 10000;
```
zcei: `Number.MAX_SAFE_INTEGER` is lower than a 17 digit unix time (it has only 16 digits).
As we probably don't want to use `BigInt` here I'd just slice away the last four digits.

samples/basic-approval-workflow/src/pubnub-client.js
```diff
@@ -0,0 +1,85 @@
...
+    return false;
+  }
+
+  function getHistory() {
+    return new Promise((resolve, reject) => {
+      pubnub.history({
+        channel,
+        count: 25,
+        stringifiedTimeToken: true,
+      }, (status, res) => {
```
zcei: Why are you using the callback interface here? 
[There seems to be a Promise version as well](https://www.pubnub.com/docs/web-javascript/api-reference-storage-and-playback#history-basic-usage-using-promises)

samples/basic-approval-workflow/src/pubnub-client.js
```diff
@@ -0,0 +1,85 @@
...
+        } else {
+          resolve(res);
+        }
+      });
+    });
+  }
+
+  function publish(message) {
+    return new Promise((resolve, reject) => {
+      pubnub.publish({ channel, message }, (status, res) => {
```
zcei: Same, here, [there's a Promise version for publish as well](https://www.pubnub.com/docs/web-javascript/api-reference-publish-and-subscribe#publish-basic-usage-using-promises)

samples/basic-approval-workflow/src/pubnub-client.js
jelz: Cool, wasn't aware of it!

PR #111, opened 2019-04-04
Setup CircleCI

realityking:
This PR sets up CircleCI for the marketplace extensions. It runs the followings jobs:

- Codestyle check using prettier
- Linting using eslint
- Unit testing using jest
- Building the extensions

Deployment and code coverage will come in future PRs


.circleci/config.yml
```diff
@@ -17,6 +17,12 @@ jobs:
           <<: *cache-key
           paths:
             - ./node_modules
+      - run:
+          name: Install extension dependencies
+          command: npm run bootstrap
+      - run:
+          name: Check code style
+          command: npm run prettier-check
```
suevalov: I don't think it's needed on CI. The codebase is automatically prettified on commit and there's `lint` step to catch actual errors.

realityking: I disagree. You can always bypass the the commit check (plus it's not even setup yet 😬), this way we make sure no one accidentally does weird things and master is always in a correct state.

The check is also super fast. With the handful of files currently in the repo it's < 1 second.

PR #112, opened 2019-04-04
Housekeeping

realityking:
- Adds a readme to the marketplace folder
- Adds more tests to the Google Analytics extension
- Enable automatic deployments

### DIFF
--- a/marketplace/Readme.md
+++ b/marketplace/Readme.md
@@ -24,7 +24,7 @@ Extensions in this directory are automatically deployed to Netlify using a domai
 `{extension-id}-{major-version}.contentfulexts.com`. The `extension-id` is the `id` property in `extension.json`, the
 `major-version` is the first segment of the `version` property in `package.json`.
 
-After adding a new extension, or updating a major version, a Contentful employee need to complete the following steps
+After adding a new extension, or updating a major version, a Contentful employee needs to complete the following steps
 after the PR is merged:
 
 * Log-in the Netlify account and navigate to the newly created site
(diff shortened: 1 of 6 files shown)


marketplace/Readme.md
```diff
@@ -0,0 +1,29 @@
...
+These extensions are production quality and ready to use out of the box.
+
+## Development
+
+Extensions should be created using `create-contentful-extension`. After copying them over the require only minimal
+modifications:
+
+* Add a `version` property to `package.json` with the value `1.0.0` or higher.
+* Add unit tests. Preferably using Jest.
+  * Modify `.babelrc` to be environments aware
```
suevalov: Can you please add more details here?

marketplace/Readme.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Deployment
+
+Extensions in this directory are automatically deployed to Netlify using a domain with the following pattern:
+`{extension-id}-{major-version}.contentfulexts.com`. The `extension-id` is the `id` property in `extension.json`, the
+`major-version` is the first segment of the `version` property in `package.json`.
+
+After adding a new extension, or updating a major version, you have to complete the following steps after the PR is merged:
+
+* Log-in the Netlify account and navigate to the newly created site
```
suevalov: I'm not sure if this information should be public. IMO, internal wiki is a better option.

suevalov:
LGTM, only two comments about docs.

marketplace/Readme.md
realityking: I think it's fine to keep this public (it's fairly evident from the deploy script) but I added a note that this has to be done by a Contentful employee.

marketplace/Readme.md
realityking: I added some more color. Can you look is this makes things clearer?

suevalov:
👍

PR #129, opened 2019-06-19
Extensions improvements

abhiaiyer91:


samples/gatsby-preview/extension.json
```diff
@@ -21,7 +21,12 @@
         "id": "webhookUrl",
         "type": "Symbol",
         "name": "The url of the webhook to rebuild your Gatsby Preview instance"
+      },
+      {
+        "id": "authToken",
```
jelz: What kind of auth token is that? Please note these tokens are visible to all space members. If it allows operations altering/removing data it shouldn't be provided as a parameter. If it's for reading only then it's fine.

abhiaiyer91: @jelz this is an auth token that allows the extension to bypass Gatsby Auth access to fetch the paths of the content. No write actions allowed. Could I get advised on a better description?

jelz: @abhiaiyer91 just wanted to verify it's safe to use this token in the fronted. All good - I've just approved the PR.

PR #135, opened 2019-06-24
Added page extension example

suevalov:


samples/page-extension-react-router/src/index.js
```diff
@@ -0,0 +1,112 @@
...
+                    Other
+                  </Tab>
+                </>
+              )}
+            />
+          </Tabs>
+          <Switch>
+            <Route path="/" exact render={() => <MainPage />} />
+            <Route path="/other" exact render={() => <OtherPage />} />
+            <Route render={() => <Page404 />} />
```
davidfateh: Non-blocking: We could make this a little more concise by just passing the React component in the `render` prop.
For example:  `<Route render={Page404} />` since we don't pass router props to any of the above components in the routes.

PR #137, opened 2019-06-25
Rename Bynder Dialog to Bynder Assets

realityking:
This is a better name for listing it on the Marketplace


realityking:
@lehnerm Marketplace PR is still WIP. This was the time to do this change :)

PR #140, opened 2019-06-26
Add Typeform UI extension

realityking:
One more extension 🎉

### DIFF
--- a/marketplace/typeform/package.json
+++ b/marketplace/typeform/package.json
@@ -1,0 +1,48 @@
+{
+  "name": "typeform",
+  "version": "1.0.0",
+  "private": true,
+  "devDependencies": {
+    "@babel/core": "7.4.5",
+    "@babel/plugin-proposal-class-properties": "7.4.4",
+    "@babel/plugin-transform-runtime": "7.4.4",
+    "@babel/preset-env": "7.4.5",
+    "@babel/preset-react": "7.0.0",
+    "@contentful/contentful-extension-scripts": "0.12.0",
+    "contentful-cli": "^0.33.2",
+    "react-test-renderer": "^16.8.6"
+  },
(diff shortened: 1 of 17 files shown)


realityking:
This is ready for review now. The previous build failure was due to CircleCI running out of memory while building the extensions. I've reduced concurrency to 2 to fix this.

suevalov:
@realityking Can you show me and @floelhoeffel what does the extension do?

realityking:
It's essentially the same functionality as our Marketo extension (except it's much simpler because the Typeform API makes things easier) and using Forma36 for the look & feel. 

The whole things looks like this:

Screen Shot 2019-06-27 at 17 48 08

Simply a select element populated with all your Typeforms.

Note I've build this to be used with a Personal Access Token so that an admin can set up for the entire team instead of using an oAuth app where every user would have to authenticate themselves.

marketplace/typeform/package.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "scripts": {
+    "configure": "contentful space use && contentful space environment use",
+    "start": "contentful-extension-scripts start",
+    "build": "contentful-extension-scripts build",
+    "deploy": "npm run build && contentful extension update --force",
+    "login": "contentful login",
+    "logout": "contentful logout",
+    "help": "contentful-extension-scripts help",
+    "test": "TZ=UTZ jest test/*.test.js",
```
davidfateh: Should this be `TZ=UTC`? 

(same comment for the line below)

realityking: Good catch. Fixed!

realityking:
Updated :)

I left the `testEnvironment` as node as that will make the tests run a little faster.

PR #142, opened 2019-06-26
Update gatsby extension to open preview url when there is no content type

rizchelwood:


suevalov:
Yep, @rizchelwood. Please, do `npm run build` before commiting it. I'll add a precommit hook soon to prevent this error.

PR #147, opened 2019-06-28
Netlify Build Button UIE

davidfateh:
## Overview
This is the Netlify Build Button UIE. You can connect it to Netlify to build your apps.

Screenshot 2019-06-28 at 09 58 21

## Todos
- [ ] Tests
 - [ ] Confirm UI Design is good


marketplace/netlify-build/src/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <link type="text/css" rel="stylesheet" href="./index.css" />
+  </head>
+  <body>
+    <div id="root"></div>
+    <script type="text/javascript" src="https://cdn.pubnub.com/sdk/javascript/pubnub.4.21.7.min.js"></script>
```
suevalov: can we import it as usual npm dependency?

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,53 @@
...
+  "name": "Netlify Build and Preview",
+  "src": "http://localhost:1234",
+  "sidebar": true,
+  "parameters": {
+    "installation": [
+      {
+        "id": "publishKey",
+        "type": "Symbol",
+        "name": "PubNub publish key",
+        "description": "A public PubNub publish key to publish update events to Netlify",
```
jelz: We don't "publish update events to Netlify", it's for notifying other users in the space.

I'd say "A public PubNub publish key to emit update events"

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,53 @@
...
+        "required": true
+      },
+      {
+        "id": "subscribeKey",
+        "type": "Symbol",
+        "name": "PubNub subscribe key",
+        "description": "A public PubNub subscribe key to listen for update events from Netlify",
```
jelz: Drop "from Netlify"

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,53 @@
...
+        "required": true
+      },
+      {
+        "id": "channels",
+        "type": "Symbol",
+        "name": "Comma seperated Netlify channel list (Max 3)",
```
jelz: These are PubNub channels. Would you capitalize "max"?

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,53 @@
...
+        "description": "Example: my-app-abc,my-second-app-def,my-third-app-ghi"
+      },
+      {
+        "id": "netlifySiteIds",
+        "type": "Symbol",
+        "name": "Comma seperated Netlify Id list (Max 3)",
```
jelz: "Netlify Site ID list"

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,53 @@
...
+      },
+      {
+        "id": "subscribeKey",
+        "type": "Symbol",
+        "name": "PubNub subscribe key",
+        "description": "A public PubNub subscribe key to listen for update events from Netlify",
+        "required": true
+      },
+      {
+        "id": "channels",
```
jelz: Shouldn't we `required` this and the following?

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,53 @@
...
+        "id": "netlifySiteIds",
+        "type": "Symbol",
+        "name": "Comma seperated Netlify Id list (Max 3)",
+        "description": "Example: 14234e3,143248,35233"
+      },
+      {
+        "id": "buildHookUrls",
+        "type": "Symbol",
+        "name": "Comma seperated Netlify build hook URLs (Max 3)",
+        "description": "https://api.netlify.com/build_hooks/{site-id}"
```
jelz: "Example: https://api.netlify.com/build_hooks/xyz,https://api.netlify.com/build_hooks/abc". There are no site IDs in URLs.

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,53 @@
...
+      },
+      {
+        "id": "names",
+        "type": "Symbol",
+        "name": "Comma seperated Website name list (Max 3)",
+        "description": "options,My Site"
```
jelz: "Example:"

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,53 @@
...
+      },
+      {
+        "id": "netlifySiteUrls",
+        "type": "Symbol",
+        "name": "Comma seperated Netlify site URLs (Max 3)"
```
jelz: description missing

marketplace/netlify-build/src/NetlifySideBarBuildButton.js
```diff
@@ -0,0 +1,131 @@
...
+    users: PropTypes.arrayOf(PropTypes.object.isRequired).isRequired,
+    userId: PropTypes.string.isRequired,
+    publishKey: PropTypes.string.isRequired,
+    subscribeKey: PropTypes.string.isRequired
+  };
+
+  state = { history: [] };
+
+  async componentDidMount() {
+    this.createSub();
```
jelz: `createPubSub`

marketplace/netlify-build/src/NetlifySideBarBuildButton.js
```diff
@@ -0,0 +1,131 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button, ValidationMessage } from '@contentful/forma-36-react-components';
+import { normalizeMessage, isOutOfOrder, isDuplicate, messageToState } from './message-processor';
+
+import '@contentful/forma-36-react-components/dist/styles.css';
+import styles from './styles';
+
+const EVENT_TRIGGERED = 'triggered';
```
jelz: Shouldn't these values come from a centralized constant list?

marketplace/netlify-build/src/NetlifySideBarBuildButton.js
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  build = async () => {
+    this.pubsub.publish({
+      contentful: true,
+      event: EVENT_TRIGGERED,
+      userId: this.props.userId
+    });
+
+    const res = await fetch(this.props.site.buildHookUrl, { method: 'POST' });
```
jelz: We could polyfill `fetch`

marketplace/netlify-build/src/index.html
jelz: Either install from npm or load from the CDN lazily

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import { init } from 'contentful-ui-extensions-sdk';
+import { createPubSub } from './pubnub-client';
+import { Select, Option } from '@contentful/forma-36-react-components';
+import NeflifySideBarBuildButton from './NetlifySideBarBuildButton';
+
+function createSiteStructure(parameters = {}) {
+  const buildHooks = parameters.buildHookUrls.split(',');
```
jelz: I'd always default to an empty string `(parameters.buildHookUrls || '').split(',')`

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+  const channels = parameters.channels.split(',');
+  const siteIds = parameters.netlifySiteIds.split(',');
+  const names = parameters.names.split(',');
+  const urls = parameters.netlifySiteUrls.split(',');
+
+  const sites = [];
+
+  for (let i = 0; i < buildHooks.length; i++) {
```
jelz: `reduce` here

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,85 @@
...
+    const { items } = await this.props.sdk.space.getUsers();
+    this.setState({ users: items });
+  }
+
+  selectSite = e => {
+    this.setState({ selectedSiteIndex: Number(e.target.value) });
+  };
+
+  render() {
+    if (!this.state.users.length) {
```
jelz: why?

marketplace/netlify-build/src/message-processor.js
```diff
@@ -0,0 +1,140 @@
+import { format, distanceInWordsToNow } from 'date-fns';
+
+export const EVENT_TRIGGERED = 'triggered';
```
jelz: Let's introduce `constants.js` will all these things

marketplace/netlify-build/src/pubnub-client.js
```diff
@@ -0,0 +1,110 @@
...
+        }
+      }
+    };
+
+    state.instance.addListener(state.mainListener);
+    state.instance.subscribe({ channels });
+  }
+
+  function publish(message) {
+    return new Promise((resolve, reject) => {
```
jelz: I've learned recently that if you don't provide callbacks to PubNub SDK methods they will return promises. So no need to wrap.

marketplace/netlify-build/test/index.spec.js
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+import TestRenderer from 'react-test-renderer';
+import NetlifyExtension from '../src/index.js';
+
+const pubsubMock = jest.fn();
+
+describe('NetlifyExtension', () => {
+  it('should render null if there are no users', () => {
```
jelz: why's that?

marketplace/netlify-build/extension.json
davidfateh: I would but since we already show `max 3` in the main Netlify configuration page, I will change it to lowercase to keep consistency.

marketplace/netlify-build/src/index.js
davidfateh: Good call, will change.

marketplace/netlify-build/src/index.html
davidfateh: Done

marketplace/netlify-build/extension.json
davidfateh: Done.

marketplace/netlify-build/src/NetlifySideBarBuildButton.js
davidfateh: Done.

marketplace/netlify-build/src/NetlifySideBarBuildButton.js
davidfateh: Done.

marketplace/netlify-build/src/index.js
davidfateh: I did some further investigation and couldn't find a good reason so I'm removing it.

marketplace/netlify-build/test/index.spec.js
davidfateh: I removed this logic and also this test.

marketplace/netlify-build/src/pubnub-client.js
davidfateh: I will give that a shot.

marketplace/netlify-build/src/index.js
davidfateh: Done.

marketplace/netlify-build/src/message-processor.js
davidfateh: Done.

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,59 @@
...
+        "type": "Symbol",
+        "name": "Comma seperated Netlify build hook URLs (max 3)",
+        "description": "Example: https://api.netlify.com/build_hooks/xyz,https://api.netlify.com/build_hooks/abc",
+        "required": true
+      },
+      {
+        "id": "names",
+        "type": "Symbol",
+        "name": "Comma seperated Website name list (max 3)",
+        "description": "Example: My Site",
```
jelz: I would always do comma separated values in examples

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,81 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import PropTypes from 'prop-types';
+import { init } from 'contentful-ui-extensions-sdk';
+import { createPubSub } from './pubnub-client';
+import { Select, Option } from '@contentful/forma-36-react-components';
+import NeflifySideBarBuildButton from './NetlifySideBarBuildButton';
+
+function createSiteStructure(parameters = {}) {
+  const buildHooks = (parameters.buildHookUrls || '').split(',');
```
jelz: I would do something like this:

```js
const readCsvParam = csv => (csv || '')
  // split by comma
  .split(',')
  // remove whitespace around the value
  .map(val => val.trim())
  // filter out empty values in case of double comma `,,`
  .filter(val => val.length > 0)

const buildHooks = readCsvParam(parameters.buildHookUrls)
// ...
```

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+  const channels = (parameters.channels || '').split(',');
```
jelz: Actually channel names need to:

1. be stable
2. be globally unique per installation per site

Now I see we actually don't need it as a parameter. I would use suffixes of build hook URLs:

```js
const buildHookUrls = readCsvParam(paramters.buildHookUrls)
const channels = buildHookUrls.map(hook => hook.split('/').pop())
```

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+  };
+
+  async componentDidMount() {
+    this.props.sdk.window.startAutoResizer();
+    const { items } = await this.props.sdk.space.getUsers();
+    this.setState({ users: items });
+  }
+
+  selectSite = e => {
+    this.setState({ selectedSiteIndex: Number(e.target.value) });
```
jelz: I wouldn't use `new`-less constructors, just do `parseInt(value, 10)`

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+          users={this.state.users}
+          createPubSub={this.props.createPubSub}
+          userId={this.props.sdk.user.sys.id}
+          site={this.state.sites[this.state.selectedSiteIndex]}
+          publishKey={this.props.sdk.parameters.installation.publishKey}
+          subscribeKey={this.props.sdk.parameters.installation.subscribeKey}
+        />
+        <Select onChange={this.selectSite}>
+          {this.state.sites.map((site, index) => (
+            <Option key={site.netlifySiteId} value={String(index)}>
```
jelz: As above, use:

```js
<Option value ={`${index}`}>
```

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+              {site.name} (Netlify app)
```
jelz: We're not using content preview anymore, it'll get rendered in Netlify App sidebar section so we can drop the "(Netlify app)"

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,81 @@
...
+          ))}
+        </Select>
+      </>
+    );
+  }
+}
+
+init(sdk => {
+  ReactDOM.render(
+    <NetlifyExtension sdk={sdk} createPubSub={createPubSub} />,
```
jelz: Why do you pass `createPubSub` as a prop?

marketplace/netlify-build/src/pubnub-client.js
```diff
@@ -0,0 +1,86 @@
+import PubNub from 'pubnub';
+
+function isObject(value) {
+  var type = typeof value;
```
jelz: We've got `var` and `!=` in this function. I also don't understand why we allow functions. I also believe this code will allow an array.

I'd do:

```js
const isObject = val => typeof val === 'object' && val !== null && !Array.isArray(val)
```

marketplace/netlify-build/src/pubnub-client.js
```diff
@@ -0,0 +1,86 @@
...
+  function publish(message) {
+    return state.instance.publish({
+      message,
+      channel,
+      storeInHistory: true
+    });
+  }
+
+  function getHistory(count = 25) {
+    return state.instance.history({ channel, count, stringifiedTimeToken: true }).then(res =>
```
jelz: How about making it `async`?

marketplace/netlify-build/src/styles.js
```diff
@@ -0,0 +1,32 @@
...
+    background: tokens.colorBlueDark,
+    color: tokens.colorWhite,
+    padding: tokens.spacing2Xs,
+    fontSize: '0.65rem',
+    lineHeight: '0.65rem',
+    borderRadius: '3px',
+    textTransform: 'uppercase'
+  }),
+  body: css({
+    fontFamily: 'Avenir Next W01'
```
jelz: @suevalov do you happen to know what's the current consensus on fonts we should use in extensions?

suevalov: @jelz we're going to proceed with `system` font in the web app. As far as I know the swap is planned for Q3.

marketplace/netlify-build/src/index.js
davidfateh: Any specific reason for that?

marketplace/netlify-build/src/index.js
davidfateh: At first it was to be able to inject props into the main app which would make mocking these easier. But as I work through the more in-depth tests, I can just use `jest` to mock these imports no problem anyway. So I can remove it and just have it be an internal thing now.

marketplace/netlify-build/src/pubnub-client.js
davidfateh: Yep, I can change these. Originally I copied them from `lodash` so I wouldn't need to import the whole package just to get some very minor functionality.

marketplace/netlify-build/src/styles.js
davidfateh: Should I leave this as-is for now or remove it?

suevalov: I'd leave `Avenir` for now and replace fonts in all extension when a swap happens

marketplace/netlify-build/src/index.js
davidfateh: I haven't actually implemented a button to send a user to the preview of the site. It's still possible to see your preview using the content preview button. Should we add another button to this UIE that actually opens the preview in a new tab as well?

marketplace/netlify-build/src/index.js
davidfateh: Good call, will change.

davidfateh:
> Some more tests would be nice :)

Still working on a more in-depth test. Basically need to mock out ALOT to get things actually in a place where we can test functionality.

marketplace/netlify-build/src/index.js
jelz: It just confuses me when I read the code - I'm not sure what the behavior is, I'm not aware of edge cases. Also with `parseInt` I see I'm dealing with integers. For different class of errors I'd expect constructors to be called with `new` and here we just use it as a function.

marketplace/netlify-build/src/index.js
jelz: Yes we should do it. We don't want to use content preview. It doesn't support environments. With this extension you can actually get environment-aware website preview. I'd even consider replacing the default preview sidebar widget with this one in the installation process.

marketplace/netlify-build/src/index.js
davidfateh: Ok thanks for the explanation. In this specific use case the functionality is identical. However, I think readability will trump semantics ;) I'll change it.

marketplace/netlify-build/src/index.js
davidfateh: I want to make sure I'm following 100% here. Are you saying that we can stop creating the channel names in the main UI code and just have this UIE create the channel names here?

marketplace/netlify-build/src/index.js
davidfateh: Ok that makes sense. I'll add in the functionality now.

suevalov:
@davidfateh can you please rebase with the latest master and update to `create-extensions-scripts@0.12.0`?

marketplace/netlify-build/src/index.js
davidfateh: Followed up offline, this is the case.

marketplace/netlify-build/src/NetlifySideBarBuildButton.js
```diff
@@ -0,0 +1,131 @@
...
+  };
+
+  state = { history: [] };
+
+  async componentDidMount() {
+    this.createPubSub();
+  }
+
+  componentDidUpdate(prevProps) {
+    // on initial render we may not get a user array
```
jelz: why?

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,94 @@
...
+  const siteIds = readCsvParam(parameters.netlifySiteIds);
+  const names = readCsvParam(parameters.names);
+  const urls = readCsvParam(parameters.netlifySiteUrls);
+
+  return buildHooks.reduce(
+    (acc, _, i) =>
+      acc.concat([
+        {
+          buildHookUrl: buildHooks[i],
+          channel: buildHooks[i].split('/').pop(),
```
jelz: we could add a comment for what we do here

marketplace/netlify-build/src/pubnub-client.js
```diff
@@ -0,0 +1,84 @@
...
+  async function publish(message) {
+    return await state.instance.publish({
+      message,
+      channel,
+      storeInHistory: true
+    });
+  }
+
+  async function getHistory(count = 25) {
+    return await state.instance.history({ channel, count, stringifiedTimeToken: true }).then(res =>
```
jelz:
```js
const res = await history()
return (res.messages || []).map(/* .... */)
```

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,94 @@
...
+
+  state = {
+    users: [],
+    sites: createSiteStructure(this.props.sdk.parameters.installation),
+    selectedSiteIndex: 0
+  };
+
+  async componentDidMount() {
+    this.props.sdk.window.startAutoResizer();
+    const { items } = await this.props.sdk.space.getUsers();
```
jelz: Ah I see why they are not defined right away. I'd do a loading state, with skeleton preferably.

davidfateh: I investigated this originally and found that really the only reason to have `users` is to show a user name for the messaging that happens when a build is triggered (the text that shows up under the build button). In this case, you can actually still use the application without needing to have `users`. What ends up happening is usually this loads so fast that you don't notice anything. But in the off chance that this doesn't load right away, you still see 99% of the application including the build message. It just doesn't include a username. I don't think it's worth it to make the whole thing have a loading screen to wait for this small detail.

davidfateh:
@suevalov when you get a chance, could you please review the testing setup? I've followed your instructions and I think I have everything up to date.

marketplace/netlify-build/src/index.js
davidfateh: This being said, I'll add a test to ensure we keep this functionality the same.

marketplace/netlify-build/src/index.js
davidfateh: Added.

marketplace/netlify-build/.babelrc
```diff
@@ -0,0 +1,48 @@
...
+      "@babel/plugin-transform-runtime",
+      {
+        "corejs": false,
+        "helpers": false,
+        "regenerator": true
+      }
+    ]
+  ],
+  "env": {
+    "test": {
```
suevalov: not sure if we need it with a new jest setup

suevalov:
@davidfateh looks good

marketplace/netlify-build/.babelrc
davidfateh: I'll take a look

marketplace/netlify-build/src/index.js
jelz: Ok, makes sense. Just make sure it looks ok if we don't have users yet

marketplace/netlify-build/.babelrc
davidfateh: Ok, removed.

marketplace/netlify-build/src/index.js
davidfateh: I think it looks good. The differences are like this:

no users:
>  Triggered about 4 hours ago at 5:27:05 PM.

users:
> Triggered by David Fateh about 4 hours ago at 5:27:05 PM.

That's the only UI difference.

davidfateh:
I've added a README to this extension.

marketplace/netlify-build/README.md
```diff
@@ -0,0 +1,25 @@
+# Netlify Build and Preview
+
+This sidebar extension allows you to build and preview up to 3 Netlify sites.
```
jelz: I'd be more descriptive, add something similar to: "...directly from your entry editor. It's also possible to navigate to the site with a single click"

marketplace/netlify-build/README.md
```diff
@@ -0,0 +1,25 @@
...
+![Netlify Sidebar UI Sample](./assets/netlify-triggering.gif)
+
+## Requirements
++ A Netlify site
++ Netlify webhooks setup, [see their documentation](https://www.netlify.com/docs/webhooks/)
+
+## Installation
+
+Navigate to the Extensions page to install the extension and provide the
+correct credentials into the required fields then click "Save".
```
jelz: "correct configuration: build hook URL, site ID, URL and its name. To provide multiple configurations separate values with a comma."

marketplace/netlify-build/README.md
```diff
@@ -0,0 +1,25 @@
+# Netlify Build and Preview
+
+This sidebar extension allows you to build and preview up to 3 Netlify sites.
+
+This is an alternative approach to using a webhook to trigger a build. The advantage
+is that the editor gets control on when a build is executed instead of responding
+to automatic events such as `Entry.publish`.
+
+![Netlify Sidebar UI Sample](./assets/netlify-triggering.gif)
```
jelz: Can we have the full flow in the gif? "Triggering, building, last built by..."

marketplace/netlify-build/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Requirements
++ A Netlify site
++ Netlify webhooks setup, [see their documentation](https://www.netlify.com/docs/webhooks/)
+
+## Installation
+
+Navigate to the Extensions page to install the extension and provide the
```
jelz: Explain how to install. Provide a deeplink. Use Netlify-deployed install URL.

marketplace/netlify-build/README.md
```diff
@@ -0,0 +1,25 @@
...
+correct credentials into the required fields then click "Save".
+
+![Netlify Installation Config](./assets/setup.png)
```
jelz: You show `localhost` in this screenshot. It's OK but would be better if we'd show the actual deployment URL.

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,38 @@
...
+  "id": "netlify-button",
+  "name": "Netlify Build and Preview",
+  "src": "http://localhost:1234",
+  "sidebar": true,
+  "parameters": {
+    "installation": [
+      {
+        "id": "netlifySiteIds",
+        "type": "Symbol",
+        "name": "Comma seperated Netlify Site Id list (max 3)",
```
jelz: Capitalize "ID"

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,38 @@
...
+        "description": "Example: 14234e3,143248,35233",
```
jelz: Provide values that resemble Netlify site IDs

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,38 @@
...
+        "required": true
+      },
+      {
+        "id": "buildHookUrls",
+        "type": "Symbol",
+        "name": "Comma seperated Netlify build hook URLs (max 3)",
+        "description": "Example: https://api.netlify.com/build_hooks/xyz,https://api.netlify.com/build_hooks/abc",
```
jelz: Replace xyz, abc with values that resemble Netlify Build hook URLs

marketplace/netlify-build/extension.json
```diff
@@ -0,0 +1,38 @@
...
+        "type": "Symbol",
+        "name": "Comma seperated Website name list (max 3)",
+        "description": "Example: My Site,My Other Site",
+        "required": true
+      },
+      {
+        "id": "netlifySiteUrls",
+        "type": "Symbol",
+        "name": "Comma seperated Netlify site URLs (max 3)",
+        "description": "Example: https://firefly-12345.netlify.com",
```
jelz: In all examples provide the same number of values, I think 2 is fine

marketplace/netlify-build/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "netlify-button",
+  "version": "0.1.0",
```
jelz: Let's bump to 1.0

marketplace/netlify-build/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "netlify-button",
```
jelz: How about `netlify-build-and-preview`?

marketplace/netlify-build/src/NetlifySideBarBuildButton.js
```diff
@@ -0,0 +1,131 @@
...
+    const hasNewUsers = !prevProps.users.length && this.props.users.length;
+
+    if (hasNewUsers || (this.props.site !== prevProps.site && this.pubsub)) {
```
jelz: Do we need to stop and start if we get new users?

marketplace/netlify-build/src/NetlifySideBarBuildButton.js
```diff
@@ -0,0 +1,131 @@
...
+      this.pubsub.stop();
+      this.createPubSub();
+    }
+  }
+
+  createPubSub = async () => {
+    const { site } = this.props;
+
+    if (!site.channel || !site.netlifySiteId || !site.buildHookUrl) {
```
jelz: How about name?

marketplace/netlify-build/src/index.css
```diff
@@ -0,0 +1,11 @@
+html,
+body,
+div {
+  margin: 0;
+  padding: 0;
+  border: 0;
+  font-size: 100%;
+  font: inherit;
+  vertical-align: baseline;
+  font-family: "Avenir Next W01", sans-serif;
```
jelz: So the decision was we keep Avenir?

marketplace/netlify-build/src/index.js
```diff
@@ -0,0 +1,96 @@
...
+    this.setState({ selectedSiteIndex: parseInt(e.target.value, 10) });
+  };
+
+  render() {
+    return (
+      <>
+        <NeflifySideBarBuildButton
+          users={this.state.users}
+          userId={this.props.sdk.user.sys.id}
+          site={this.state.sites[this.state.selectedSiteIndex]}
```
jelz: You could `const site = this.state.sites[this.state.selectedSiteIndex]` on the top of the method

marketplace/netlify-build/src/pubnub-client.js
```diff
@@ -0,0 +1,83 @@
+import PubNub from 'pubnub';
+import { PUBLISH_KEY, SUBSCRIBE_KEY } from './contstants';
+
+const isObject = val => typeof val === 'object' && !Array.isArray(val);
```
jelz:
```
> typeof null
'object'
```

I would add `&& val !== null`

marketplace/netlify-build/src/styles.js
```diff
@@ -0,0 +1,46 @@
...
```
jelz: This seems to repeat what we've got in `index.css`. Can we merge the two?

marketplace/netlify-build/test/index.spec.js
```diff
@@ -0,0 +1,232 @@
...
+    instance: {},
+    installation: {
+      names: 'upbeat,friendly',
+      channels:
+        'c-wj2QelNg1OvkkWDF-5133ac93-7fc0-463b-96d0-bb799e7fcb07,c-wj2QelNg1OvkkWDF-119493e8-f2a3-4771-8b3c-f402b5966e84',
+      publishKey: 'pub-c-a99421b9-4f21-467b-ac0c-d0292824e8e1',
+      subscribeKey: 'sub-c-3992b1ae-f7c9-11e8-adf7-5a5b31762c0f',
+      buildHookUrls:
+        'https://api.netlify.com/build_hooks/5d19c2c1b359da07eca189af,https://api.netlify.com/build_hooks/5d19c2c1b359da2f7da18998',
+      netlifySiteIds: '5133ac93-7fc0-463b-96d0-bb799e7fcb07,119493e8-f2a3-4771-8b3c-f402b5966e84',
```
jelz: Are these real? We can keep them but make sure you deleted Netlify sites.

marketplace/netlify-build/src/index.css
davidfateh: Yes, for now.

marketplace/netlify-build/src/pubnub-client.js
davidfateh: Ah good catch.

marketplace/netlify-build/test/index.spec.js
davidfateh: They are real and I own them. They will be deleted from my personal Netlify once I'm done so no worries there.

marketplace/netlify-build/src/styles.js
davidfateh: I'm going to get rid of this value in favor of the other one as it covers more cases.

marketplace/netlify-build/src/NetlifySideBarBuildButton.js
davidfateh: Short answer: No.

But that would mean refactoring how the messaging system works. If you look on line 46, you'll see that there is a preconfigured message normalizer function that will get passed into the main listener which subscribes to pubnub messages. I could refactor that to be able to dynamically add in user info if some new info comes along but I figured it's just easier to stop it and start it up again. It's the same logic path we take if the `site` prop were to change and it seems to be reasonable there.

marketplace/netlify-build/package.json
davidfateh: Works for me!

marketplace/netlify-build/README.md
davidfateh: Are you saying I should provide instructions on how to install a Contentful extension into the space? It seems like none of the other marketplace extensions have that in the README, I assumed because it would be taken care of by the marketplace UI itself.

marketplace/netlify-build/README.md
davidfateh: Added in more detail.

marketplace/netlify-build/README.md
davidfateh: Added in more detail.

marketplace/netlify-build/src/index.js
davidfateh: Done.

marketplace/netlify-build/src/pubnub-client.js
davidfateh: Done.

marketplace/netlify-build/src/styles.js
davidfateh: Done.

marketplace/netlify-build/extension.json
davidfateh: Done.

marketplace/netlify-build/extension.json
davidfateh: Done.

marketplace/netlify-build/extension.json
davidfateh: Done.

marketplace/netlify-build/package.json
davidfateh: Done

marketplace/netlify-build/package.json
davidfateh: Done.

marketplace/netlify-build/src/NetlifySideBarBuildButton.js
davidfateh: Done.

marketplace/netlify-build/extension.json
davidfateh: Done.

marketplace/netlify-build/README.md
davidfateh: I'm adding more installation steps with a guide. Should be up soon.

marketplace/netlify-build/README.md
davidfateh: Done.

marketplace/netlify-build/README.md
davidfateh: Done.

davidfateh:
I added in some detailed installation instructions. Could use some eyes on it.

PR #155, opened 2019-07-02
Fix the translation fallback when using Bynder

zuuperman:


marketplace/bynder-assets/src/Field.js
```diff
@@ -25,7 +25,12 @@ function Field({ sdk }) {
   }, []);
 
   useEffect(() => {
-    sdk.field.setValue(assets);
+    if (assets.length) {
+      sdk.field.setValue(assets);
+    }
+    else {
+      sdk.field.setValue('');
```
jelz: @suevalov @zuuperman I think it would be better to call `sdk.field.removeValue()` here

PR #157, opened 2019-07-03
Apply new eslint-config and fix all errors

suevalov:


scripts/.eslintrc
```diff
@@ -2,15 +2,10 @@
   "env": {
     "node": true
   },
-  "extends": [
-    "../.eslintrc",
```
realityking: @suevalov I might be missing something, but didn't we lose all the default eslint and promise plugins checks here?

PR #160, opened 2019-07-09
Updates Readmes and Marketplace descriptions for some Marketplace extensions

realityking:


marketplace/Readme.md
```diff
@@ -27,7 +32,7 @@ Extensions in this directory are automatically deployed to Netlify using a domai
 After adding a new extension, or updating a major version, a Contentful employee needs to complete the following steps
 after the PR is merged:
 
-* Log-in the Netlify account and navigate to the newly created site
-  * Create the DNS record for the domain
-  * Enable HTTPS for the custom domain
-* Add the extension to the Marketplace
+- Log-in the Netlify account and navigate to the newly created site
```
jelz: Does it make sense to move it to the wiki instead of keeping in a public repo?

realityking: It's a fair question. My thinking was to keep it here (it's prefaced with "a Contentful employee needs to") because the wiki page might be hard to find if you don't know about. But no strong opinion from me here.

floelhoeffel: maybe link to a readme in doc-app (which is private)?

PR #162, opened 2019-07-10
Testing Optimizely

davidfateh:
I've added unit tests to the top level components and the entry point logic. I believe further testing should be done on the `App` itself but if that should be in the unit test world is an open question. I personally think it will be easier to test at the integration level if we don't have to mock out as many things. Let me know your thoughts.

### DIFF
--- a/marketplace/optimizely/package.json
+++ b/marketplace/optimizely/package.json
@@ -17,6 +17,7 @@
     "@contentful/forma-36-fcss": "0.0.20",
     "@contentful/forma-36-react-components": "^3.11.4",
     "@contentful/forma-36-tokens": "^0.3.0",
+    "@testing-library/react": "^8.0.4",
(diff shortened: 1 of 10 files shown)


marketplace/optimizely/package.json
```diff
@@ -40,5 +42,10 @@
     "last 5 Chrome version",
     "> 1%",
     "not ie <= 11"
-  ]
+  ],
+  "jest": {
+    "moduleNameMapper": {
+      "\\.(css|less)$": "identity-obj-proxy"
```
davidfateh: @suevalov I found that I still needed this, otherwise CSS would fail. Potentially we need to change this?

https://github.com/contentful/create-contentful-extension/blob/master/packages/contentful-extension-scripts/scripts/utils/createJestConfig.js#L15

PR #166, opened 2019-07-22
Updated marketplace copy for Commerce Layer extension. 

Shy:


marketplace/commercelayer-sku/MARKETPLACE.md
```diff
@@ -0,0 +1,29 @@
+# Contentful UI Extension
```
realityking: The Marketplace files don't have a header, the header is already rendered by the doc-app

PR #168, opened 2019-07-24
Netlify app

jelz:


marketplace/netlify-build/src/netlify-integration.js
```diff
@@ -0,0 +1,139 @@
...
+
+  // ...remove Netlify hooks for it.
+  let netlifyHookRemovalPromises = Promise.resolve();
+
+  if (Array.isArray(netlifyHookIds)) {
+    const validNetlifyHookIds = netlifyHookIds.filter(id => {
+      return typeof id === 'string' && id.length > 0;
+    });
+
+    netlifyHookRemovalPromises = Promise.all(
```
davidfateh: Minor: I don't think it's necessary to do a `Promise.all` here since it concats with the array on line 134. It might be better to remove it here since the logic below handles it in a `try...catch` where this one will throw.

marketplace/netlify-build/src/target-state.js
```diff
@@ -0,0 +1,19 @@
...
+
+  return ctIds.filter(id => !!get(eiState, [id, 'sidebar']));
+}
+
+export function enabledContentTypesToTargetState(contentTypes, enabledContentTypes) {
+  return {
+    EditorInterface: contentTypes.reduce((acc, [ctId]) => {
+      return {
+        ...acc,
+        [ctId]: enabledContentTypes.includes(ctId) ? { sidebar: { position: 2 } } : {}
```
davidfateh: Should we be hardcoding position 2 here? Maybe this should be an installation option so it doesn't mess with peoples' sidebars if they don't want.

marketplace/netlify-build/src/netlify-app-config.js
```diff
@@ -0,0 +1,254 @@
...
+export default class NetlifyAppConfig extends React.Component {
+  static propTypes = {
+    sdk: PropTypes.object.isRequired
+  };
+
+  async componentDidMount() {
+    this.init();
+  }
+
+  componenWillUnmount() {
```
davidfateh:
```suggestion
  componentWillUnmount() {
```

marketplace/netlify-build/src/netlify-app-config.js
```diff
@@ -0,0 +1,254 @@
...
+
+  init = async () => {
+    const { sdk } = this.props;
+    const { app } = sdk.platformAlpha;
+
+    app.onConfigure(this.onAppConfigure);
+
+    const [parameters, currentState, contentTypesResponse] = await Promise.all([
+      app.getParameters(),
+      app.getCurrentState(),
```
davidfateh: Non-blocking: I'm starting to wonder if these values should actually get passed into the callback `app.onConfigure` takes for ease-of-use.

marketplace/netlify-build/src/netlify-app-config.js
```diff
@@ -0,0 +1,254 @@
...
+
+  render() {
+    return (
+      <Workbench>
+        <Workbench.Content>
+          <div className={styles.section}>
+            <h3>About</h3>
+            <p>
+              With this app developers can do a very quick set up, authors can control when the
+              static pages are created and see the current status of the build process.
```
davidfateh: I think this copy could use some love. Something about it seems too casual.

marketplace/netlify-build/src/netlify-app-config.js
```diff
@@ -0,0 +1,254 @@
...
+                </p>
+                <NetlifyConfigEditor
+                  disabled={!this.state.token}
+                  siteConfigs={this.state.config.sites}
+                  netlifySites={this.state.netlifySites}
+                  onSiteConfigsChange={this.onSiteConfigsChange}
+                />
+              </div>
+              <div className={styles.section}>
+                <h3>Enable for content types</h3>
```
davidfateh:
```suggestion
                <h3>Enable Netlify builds for content types</h3>
```

marketplace/netlify-build/src/netlify-app-config.js
davidfateh: I should clarify: Not saying to remove them from the app API, just wondering if we can also pass them into the callback as arguments that way you can have a sync callback if those are the only values you need.

marketplace/netlify-build/src/netlify-app-config.js
```diff
@@ -0,0 +1,254 @@
...
+    this.stopPolling();
+
+    this.cancelTicketPolling = NetlifyClient.getAccessTokenWithTicket(
+      this.state.ticketId,
+      (err, token) => {
+        if (err) {
+          this.notifyError(err, 'Failed to connect with Netlify. Try again!');
+        } else if (token) {
+          this.initNetlifyConnection(token);
+        }
```
davidfateh: What happens in the `else` case here where there is no `err` or `token`? Should we alert the user something is wrong?

marketplace/netlify-build/src/netlify-app-config.js
jelz: Computing current state can be HTTP intensive. If you don't need it but we still pass it to the callback we waste time.

marketplace/netlify-build/src/netlify-app-config.js
```diff
@@ -0,0 +1,254 @@
...
+
+    // Initially we are not connected to Netlify and we don't have
+    // a full list of Netlify sites.
+    // Here we are computing a list consisting of sites we know
+    // of so we can offer Netlify site labels even before connecting.
+    const netlifySites = config.sites
+      .filter(s => s.netlifySiteName)
+      .map(s => ({ id: s.netlifySiteId, name: s.netlifySiteName }));
+
+    const ticketId = await NetlifyClient.createTicket();
```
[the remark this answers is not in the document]
jelz: there's no information about expiry. I assume there is some but I've never run into it.

marketplace/netlify-build/src/netlify-app-config.js
```diff
@@ -0,0 +1,254 @@
...
+      this.notifyError('You must be connected to Netlify to configure the app.');
+      return false;
+    }
+
+    try {
+      const isInstalled = await this.props.sdk.platformAlpha.app.isInstalled();
+      const method = isInstalled ? 'update' : 'install';
+      const config = await NetlifyIntegration[method]({
+        config: this.state.config, // eslint-disable-line react/no-access-state-in-setstate
+        accessToken: this.state.token // eslint-disable-line react/no-access-state-in-setstate
```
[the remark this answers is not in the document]
jelz: Yes they are... I'm not sure about this rule

marketplace/netlify-build/src/netlify-app-config.js
davidfateh: Makes sense.

marketplace/netlify-build/extension.json
```diff
@@ -2,14 +2,13 @@
...
-  "sidebar": true,
   "parameters": {
     "installation": [
       {
-        "id": "netlifySiteIds",
+        "id": "notificationHookIds",
         "type": "Symbol",
-        "name": "Comma seperated Netlify Site ID list (max 3)",
-        "description": "Example: 52833133ac93-7ac0-463b-96d0-bc797e7fcb07,219463a8-f2b3-4c71-8z3c-f403b5966e80",
+        "name": "Comma separated Netlifynotification hook IDs",
```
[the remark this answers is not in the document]
jelz: Added a space, but "max 3" is not true. There are up to 3 notification hooks for each site. This value is autogenerated, I'm not going to indicate "max 9".

marketplace/netlify-build/src/netlify-app-config.js
jelz: This is handled now.

marketplace/netlify-build/src/target-state.js
jelz: This would be a good follow up but let's keep it as it is right now.

marketplace/netlify-build/src/app/netlify-integration.js
```diff
@@ -0,0 +1,139 @@
...
+  const uniqueNames = uniqBy(siteConfigs, config => config.name);
+
+  if (uniqueNames.length !== siteConfigs.length) {
+    throw makeError('Display names must be unique.');
+  }
+}
+
+function makeError(message) {
+  const err = new Error(message);
+  err.useMessage = true;
```
davidfateh: What's the reason for this?

jelz: `notifyError` in the component checks `useMessage` on errors it catches. I use it to indicate our own errors so we know we can use `err.message` for a meaningful message. I don't want to show `TypeError: cannot call method undefined of...` if for whatever reason we fail hard. If there's no `useMessage` we use a generic error message.

PR #176, opened 2019-08-05
Optimizely tests

davidfateh:


marketplace/optimizely/src/AppPage/ContentTypes.js
```diff
@@ -136,7 +136,11 @@ export default function ContentTypes({
       <Paragraph className="f36-margin-top--m">
         Select the content types for which you want to enable A/B testing.
       </Paragraph>
-      <Button buttonType="muted" className="f36-margin-top--m" onClick={() => toggleModal(true)}>
+      <Button
+        buttonType="muted"
+        className="f36-margin-top--m"
+        onClick={() => toggleModal(true)}
+        data-testId="add-content">
```
suevalov: All Forma components have special `testId` properties for this purpose and you need to use them instead.

marketplace/optimizely/test/AppPage.spec.js
```diff
@@ -1,40 +1,26 @@
...
-
-    expect(container).toMatchSnapshot();
-  });
-
-  it.skip('should render the Optimizely config when loaded', async () => {
-    const client = {
-      getProjects: () => Promise.resolve(projectData)
-    };
-
+  // There is potentially a bug with react/testing library here. The order that these tests run
```
suevalov: Please refer in the comment to the library as `@testing-library/react`. It's easier for a search.

marketplace/optimizely/test/AppPage.spec.js
```diff
@@ -52,13 +38,39 @@ describe('AppPage', () => {
...
-    await wait();
+    expect(container).toMatchSnapshot();
+
+    act(() => {
+      fireEvent.click(getByTestId('add-content'));
+    });
+
+    getByTestId('content-type-selector').click();
+
+    getByTestId('content-type-selector').firstChild.click();
```
suevalov: why do we need to click on parent first and then on its child?

marketplace/optimizely/test/AppPage.spec.js
```diff
@@ -52,13 +38,39 @@ describe('AppPage', () => {
...
+    expect(getByTestId('content-type-selector')).toMatchSnapshot();
+  });
+
+  it('should render the AppPage with Features and connect button', () => {
+    const { container } = render(<AppPage {...basicProps} />);
+
+    expect(container).toMatchSnapshot();
```
suevalov: I'm not a biggest fan of making snapshots for the whole components. As for me it's making more noise during development rather than giving a useful information. I would query important elements and match their content using `toMatchInlineSnapshot`

davidfateh: I see where you're coming from but this snapshot feels important as the whole picture. The reason is this specific instance of the `AppPage`, has two main elements: the Features list and the Optimizely oAuth button. By taking this snapshot I was able to verify both exist while other elements do not. A diff in this snapshot later on could indicate more elements are showing that potentially shouldn't be.

If I were to individually select the elements I am expecting, then I'm also ignoring any other elements that may or may not show up in the snapshot, which is exactly what I don't want to do.

marketplace/optimizely/test/AppPage.spec.js
davidfateh: This is a select element. I was hoping to mimic clicking on the select itself, then clicking on the first `option`.

Maybe I should use more testIds on the `option`s as well...

marketplace/optimizely/src/AppPage/ContentTypes.js
davidfateh: I ran into a problem when using the `testId` property. For example my code would look like this:

```
testId={testId}
```

However, the rendered attribute would be this
```
data-test-id={testId}
```

And it appears that `getByTestId` isn't smart enough to pick up attributes like `data-test-id`. It seems to only look for something like `data-testid` or `data-testId`. 

Am I missing something here?

marketplace/optimizely/test/AppPage.spec.js
davidfateh: Done here: 7b46189f64638460de60475e54bdd10bd3933e27

marketplace/optimizely/test/AppPage.spec.js
suevalov: ah, I see... just a comment would help here.

marketplace/optimizely/test/AppPage.spec.js
suevalov: Well you can select several parts in one test to verify that all elements exist in a snapshot. But I don't insist ;)

marketplace/optimizely/src/AppPage/ContentTypes.js
suevalov: you can change selector using `configure` => https://github.com/contentful/user_interface/blob/master/tools/testing/setup-tests.js#L14

davidfateh: All set: ac628cf7aa27fc33628b43e897c70519de98cdd6

marketplace/optimizely/test/AppPage.spec.js
davidfateh: Done in: b686dddba2d5fce7dab7aa98df8e5eff34002461

davidfateh:
@suevalov I decided to move it out but keep it in each file just to be explicit.

marketplace/optimizely/test/AppPage.spec.js
davidfateh: I'm going to keep it the way it is because I don't want to select several parts. That is inclusive. I want to be exclusive and have the whole snapshot signify what should be there and what shouldn't be.

PR #172, opened 2019-07-31
Moving optimizely

davidfateh:

PR #177, opened 2019-08-06
cleaning up remaining PR comments

davidfateh:
This is addressing some of the leftover comments from #172


marketplace/optimizely/src/AppPage/ContentTypes.js
```diff
@@ -131,14 +133,14 @@ export default function ContentTypes({
   };
 
   return (
-    <div className="f36-margin-top--xl">
+    <div className={styles.body}>
```
suevalov: Instead of using classes here you can just wrap all components in `Typography` and will be padded automatically

davidfateh: I changed this element to `Typography`. However, I did have to keep some spacing on the `Button` below.

suevalov: yep, right. `Button` is not a part of Typography components.

PR #180, opened 2019-08-13
Gatsby App

davidfateh:
This work adds the Gatsby Installation flow

Screenshot 2019-08-13 at 14 22 44

(disregard optimizely in the top workbench, it's just a dev thing)


marketplace/gatsby-preview/src/AppConfig.js
```diff
@@ -1,5 +1,252 @@
...
+          }, prevState.checkedContentTypes)
+        };
+      });
+    }
+  }
+
+  configureApp = async () => {
+    const { previewUrl, webhookUrl, authToken, checkedContentTypes } = this.state;
+
+    if (!previewUrl || !webhookUrl || !authToken) {
```
davidfateh: `authToken` is optional. I'll remove the check.

marketplace/gatsby-preview/src/AppConfig.js
```diff
@@ -1,5 +1,252 @@
...
+  };
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      previewUrl: '',
+      webhookUrl: '',
+      authToken: '',
+      checkedContentTypes: {}
```
davidfateh: Should have initial state to check if there are already sidebar configurations

marketplace/gatsby-preview/extension.json
```diff
@@ -7,19 +7,20 @@
       {
         "id": "previewUrl",
         "type": "Symbol",
-        "name": "The URL of your Gatsby Cloud instance",
+        "name": "The url of your Gatsby Cloud site",
```
jelz: I'd keep the "URL" capitalized: https://en.wikipedia.org/wiki/URL

marketplace/gatsby-preview/extension.json
```diff
@@ -7,19 +7,20 @@
...
-        "name": "The URL of the webhook to rebuild your Gatsby Cloud instance",
+        "name": "The url of the webhook to rebuild your Gatsby Cloud site",
         "required": true
       },
       {
         "id": "authToken",
         "type": "Symbol",
-        "name": "Optional Authentication token for private Gatsby Cloud instances"
+        "name": "Optional Authentication token for private Gatsby Cloud sites",
+        "required": true
```
jelz: If it's optional then why it's required?

marketplace/gatsby-preview/src/AppConfig.js
```diff
@@ -1,5 +1,252 @@
...
+
+  async componentDidMount() {
+    const { app } = this.props.sdk.platformAlpha;
+    app.onConfigure(this.configureApp);
+
+    const params = (await app.getParameters()) || {};
+
+    // eslint-disable-next-line react/no-did-mount-set-state
+    this.setState(prevState => ({
+      previewUrl: params.previewUrl || prevState.previewUrl,
```
jelz: Can't we just do `|| ''`? The `prevState` reference in `componentDidMount` is kind of misleading for me

marketplace/gatsby-preview/src/AppConfig.js
jelz: This you can obtain with `app.getCurrentState().EditorInterface` in `componentDidMount`

marketplace/gatsby-preview/src/AppConfig.js
```diff
@@ -1,5 +1,252 @@
...
+          }, prevState.checkedContentTypes)
+        };
+      });
+    }
+  }
+
+  configureApp = async () => {
+    const { previewUrl, webhookUrl, authToken, checkedContentTypes } = this.state;
+
+    if (!previewUrl || !webhookUrl) {
```
jelz: You could at least check if it starts with `https?://`

marketplace/gatsby-preview/src/AppConfig.js
davidfateh: Thanks! Was looking to fix this.

marketplace/gatsby-preview/src/AppConfig.js
davidfateh: Done.

marketplace/gatsby-preview/src/AppConfig.js
davidfateh: Done.

marketplace/gatsby-preview/extension.json
davidfateh: Good catch, fixed.

marketplace/gatsby-preview/extension.json
davidfateh: Done.

marketplace/gatsby-preview/src/AppConfig.js
davidfateh: Fixed.

marketplace/gatsby-preview/src/AppConfig.js
```diff
@@ -1,5 +1,286 @@
...
+  CheckboxField,
+  Pill,
+  SkeletonContainer,
+  SkeletonBodyText
+} from '@contentful/forma-36-react-components';
+import GatsbyIcon from './GatsbyIcon';
 
-export default function AppConfig() {
-  return <div>Gatsby App config screen</div>;
+const styles = {
```
jelz: Maybe extract a separate file?

marketplace/gatsby-preview/src/AppConfig.js
```diff
@@ -1,5 +1,286 @@
...
+    const [{ EditorInterface } = {}, { items }] = await Promise.all([
+      app.getCurrentState(),
+      this.props.sdk.space.getContentTypes()
+    ]);
+
+    const previouslyCheckedTypes = Object.keys(EditorInterface || {}).filter(
+      ct => EditorInterface[ct].sidebar
+    );
+
+    if (items && items.length) {
```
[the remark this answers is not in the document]
davidfateh: Old habit from being burned with performance issues. The reason I do this is to avoid setting state when I don't need it. That being said, I realize that the content type call will rarely be empty and so maybe it doesn't make sense to micro optimize like this. I'll change it.

PR #182, opened 2019-08-14
Unify DAM apps

jelz:
This PR unifies DAM (Bynder, Cloudinary) apps using a shared base package (`shared-dam-app`) so they:

- [x] Have exactly the same mechanics
- [x] Have drag and drop
- [x] Get their configuration screen rendered automatically using extension parameters

This PR doesn't include style changes.


marketplace/bynder-assets/src/index.js
```diff
@@ -1,24 +1,96 @@
...
+    parameters: { ...config },
+    width: 1400
+  });
+
+  return Array.isArray(result) ? result : [];
+}
+
+function isDisabled() {
+  return false;
+}
```
davidfateh: Is this function even needed?

marketplace/bynder-assets/src/index.js
```diff
@@ -1,24 +1,96 @@
...
+
+function validateParameters({ bynderURL }) {
+  const hasValidProtocol = bynderURL.startsWith('https://');
+  const isHTMLSafe = ['"', '<', '>'].every(unsafe => !bynderURL.includes(unsafe));
```
davidfateh: Is this something we should introduce into other apps or is this specifically due to some functionality in this app?

marketplace/cloudinary-assets/src/index.js
```diff
@@ -0,0 +1,104 @@
...
+    insertHandler: data => sdk.close(data)
+  });
+
+  instance.show();
+
+  sdk.window.updateHeight(window.outerHeight);
+}
+
+async function openDialog(sdk, currentValue, config) {
+  const maxFiles = config.maxFiles - currentValue.length;
```
davidfateh: If this value is `<= 0` should we actually disable the dialog?

marketplace/shared-dam-app/src/AppConfig/AppConfig.spec.tsx
```diff
@@ -39,23 +40,33 @@ const makeSdkMock = () => ({
   }
 });
 
+const validate = () => null; // Means no error
```
davidfateh: Is this something that needs to be changed?

marketplace/bynder-assets/src/index.js
jelz: yes, `setup` function from `shared-dam-app` expects interface expressed as `Integration` (see type definitions). There is nothing specific to the Bynder app (as oppesed to Cloudinary with its asset limit) that could disable editing so we always return `false`

marketplace/bynder-assets/src/index.js
jelz: This is specific to the implementation of this app - see above how we use `bynderURL` parameter directly in HTML.

marketplace/cloudinary-assets/src/index.js
jelz: This is done in `isDisabled`, if it returns `true` it won't be even possible to end up in this place.

marketplace/shared-dam-app/src/AppConfig/AppConfig.spec.tsx
jelz: it's a not problematic mock ;)

marketplace/bynder-assets/src/index.js
davidfateh: Ok thanks for explaining.

PR #183, opened 2019-08-20
Netlify beautify

davidfateh:
Making the Netlify App prettier.

Screenshot 2019-08-20 at 10 25 54
Screenshot 2019-08-20 at 10 26 04


lehnerm:
Looks great, two questions:
- Should we have a common component for the outer container layout or do we leave it as an individual, per app defined thing?
- I don't like the pill component for the content types - is there anything that would cover this from our design guidelines? (maybe Malin has an idea)

davidfateh:
@lehnerm 

1. We could have a common component. I do see a lot of code reuse here but at the same time we are also saying that anyone can make their installation screen unique. So while we could do a little abstraction and provide a component for the community to use, maybe it's something we should do later down the road.

2. I will ask Malin to take a look at it. So far I haven't heard anything about that aspect of the design needing work.

davidfateh:
@lehnerm Updated the checkbox field after a design review.

Screenshot 2019-08-20 at 13 54 30

lehnerm:
LGTM, Codecov seems to be unhappy with you though :)

davidfateh:
Codecov has done this before and gets ignored :P

PR #185, opened 2019-08-21
Dam fixes

jelz:
- [x] disable editing if a field is disabled
- [x] display a note if the field value cannot be managed by the app

### DIFF
--- a/marketplace/shared-dam-app/src/Editor/Field.tsx
+++ b/marketplace/shared-dam-app/src/Editor/Field.tsx
@@ -17,6 +17,7 @@ interface Props {
 
 interface State {
   value: Hash[];
+  editingDisabled: boolean;
 }
 
 const styles = {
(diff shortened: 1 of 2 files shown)


marketplace/shared-dam-app/src/Editor/Field.tsx
```diff
@@ -84,17 +93,31 @@ export default class Field extends React.Component<Props, State> {
...
 
   render = () => {
-    const currentValue = this.state.value;
-    const hasItems = currentValue.length > 0;
+    const { value, valid, editingDisabled } = this.state;
+
+    if (!valid) {
+      return (
+        <Note noteType="warning" title="Field value is incompatibile">
+          JSON object stored in this field cannot be managed with this App.
```
davidfateh:
```suggestion
          The JSON object stored in this field cannot be managed with this App.
```

PR #189, opened 2019-08-23
Add commerce tools extension

realityking:
This extension is ready for a first review. It'll need some external testing before I feel comfortable putting it on the Marketplace but it'll be easier to have folks test it if we have it hosted someplace.

I've shared credentials to test it with @jelz via 1password. I couldn't find you in 1password @giotiskl, ping me when you're added to 1password and I'll add one. Also happy to share the credentials with anyone else who would like to test it. 

Update: Invited @giotiskl as well :)

**UI issues**

- [x] Dropdown menu is cropped by iframe 
- [x] Modal controls are hidden

### DIFF
--- a/marketplace/commercetools-products/extension.json
+++ b/marketplace/commercetools-products/extension.json
@@ -1,0 +1,51 @@
+{
+  "id": "commercetools-products",
+  "name": "commercetools Products",
+  "srcdoc": "./build/index.html",
+  "fieldTypes": [
+    "Symbol", "Symbols"
+  ],
+  "parameters": {
+    "installation": [
+      {
+        "id": "projectKey",
+        "name": "commercetools Project Key",
+        "type": "Symbol",
+        "required": true
+      },
+      {
+        "id": "clientId",
+        "name": "commercetools Authentication Client ID",
+        "type": "Symbol",
+        "required": true
+      },
+      {
+        "id": "clientSecret",
+        "name": "commercetools Authentication Client Secret",
+        "type": "Symbol",
+        "required": true
+      },
+      {
+        "id": "apiUri",
+        "name": "commercetools API Endpoint",
+        "type": "Symbol",
+        "default": "https://api.commercetools.com",
+        "required": true
+      },
+      {
+        "id": "authUri",
+        "name": "commercetools AUTH API Endpoint",
+        "type": "Symbol",
+        "default": "https://auth.commercetools.com",
+        "required": true
+      },
+      {
+        "id": "locale",
+        "name": "commercetools data locale to display",
+        "type": "Symbol",
+        "default": "en",
+        "required": true
+      }
+    ]
+  }
+}
--- a/marketplace/commercetools-products/src/apollo.js
+++ b/marketplace/commercetools-products/src/apollo.js
@@ -1,0 +1,41 @@
+import SdkAuth, { TokenProvider } from '@commercetools/sdk-auth';
+import { ApolloClient } from 'apollo-client';
+import { InMemoryCache } from 'apollo-cache-inmemory';
+import { createHttpLink } from 'apollo-link-http';
+import { setContext } from 'apollo-link-context';
+
+export async function getApolloClient({ authUri, projectKey, clientId, apiUri, clientSecret }) {
+  // @todo how do I get the refreshed token?
+  const tokenProvider = new TokenProvider({
+    sdkAuth: new SdkAuth({
+      host: authUri,
+      projectKey,
+      credentials: {
+        clientId,
+        clientSecret
+      }
+    }),
+    fetchTokenInfo: sdkAuth => sdkAuth.clientCredentialsFlow()
+  });
+
+  const accessToken = await tokenProvider.getAccessToken();
+
+  const httpLink = createHttpLink({
+    uri: `${apiUri}/${projectKey}/graphql`
+  });
+
+  const authLink = setContext((_, { headers }) => {
+    // return the headers to the context so httpLink can read them
+    return {
+      headers: {
+        ...headers,
+        authorization: accessToken ? `Bearer ${accessToken}` : ''
+      }
+    };
+  });
+
+  return new ApolloClient({
+    link: authLink.concat(httpLink),
+    cache: new InMemoryCache()
+  });
+}
--- a/marketplace/commercetools-products/src/ct-product-query.graphql
+++ b/marketplace/commercetools-products/src/ct-product-query.graphql
@@ -1,0 +1,27 @@
+query productsList($skus: [String!]) {
+  products(skus: $skus) {
+    results {
+      id
+      masterData {
+        current {
+          categories {
+            nameAllLocales {
+              locale
+              value
+            }
+          }
+          nameAllLocales {
+            locale
+            value
+          }
+          masterVariant {
+            sku
+            images {
+              url
+            }
+          }
+        }
+      }
+    }
+  }
+}
--- a/marketplace/commercetools-products/src/dialog.js
+++ b/marketplace/commercetools-products/src/dialog.js
@@ -21,12 +21,18 @@ export class CommerceToolsDialog extends React.Component {
       <React.Fragment>
         <div ref={this.pickerContainerRef} className="picker-container" />
 
-        <Button onClick={this.onInsert} buttonType="positive">
-          Insert selected Products
-        </Button>
-        <Button onClick={this.onCancel} buttonType="muted">
-          Cancel
-        </Button>
+        <div className="picker-controls">
+          <Button onClick={this.onInsert} buttonType="positive" style={{ marginRight: 16 }}>
+            Insert selected Products
+          </Button>
+          <Button onClick={this.onCancel} buttonType="muted">
+            Cancel
+          </Button>
+        </div>
+        <div>
+          {/* Horrible hack for making autoresize work as expected. Autoresize doesn't like calc witchcraft */}
+          &nbsp;
+        </div>
       </React.Fragment>
     );
   }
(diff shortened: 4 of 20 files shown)


marketplace/commercetools-products/src/index.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+  </head>
+  <body>
+    <div id="root"></div>
+    <script src="https://ct-merchant-test.herokuapp.com/app.bundle.js"></script>
```
davidfateh: What's the reason behind having this external JS? Is this something we need for commerce tools to work?

marketplace/commercetools-products/src/dialog.js
```diff
@@ -0,0 +1,91 @@
...
+    super(props);
+
+    this.pickerContainerRef = React.createRef();
+    this.picker = null;
+
+    this.onCancel = this.onCancel.bind(this);
+    this.onInsert = this.onInsert.bind(this);
+  }
+
+  render() {
```
davidfateh: Style suggestion: I would reorder these methods to match all the other React components. For example: putting the render method last. It's not really a blocker, just a consistency thing.

marketplace/commercetools-products/src/field.js
```diff
@@ -0,0 +1,203 @@
...
+
+import { Button, Spinner } from '@contentful/forma-36-react-components';
+import { ApolloProvider } from '@apollo/react-components';
+import { SortableContainer } from 'react-sortable-hoc';
+import arrayMove from 'array-move';
+
+import { getApolloClient } from './apollo.js';
+import { SortableProductCard } from './product-card.js';
+
+const SortableList = SortableContainer(function SortableList({
```
giotiskl: I would suggest avoiding keeping multiple component definitions within a single file to reduce the amount of lines of code and improve readability.

marketplace/commercetools-products/src/index.js
```diff
@@ -0,0 +1,37 @@
...
+import 'regenerator-runtime/runtime';
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+import './index.css';
+import '@contentful/forma-36-react-components/dist/styles.css';
+import '@contentful/forma-36-fcss/dist/styles.css';
+import { init, locations } from 'contentful-ui-extensions-sdk';
+
+import { CommerceToolsField } from './field.js';
```
giotiskl: i'd also optionally argue in favor of removing the `.js` extension from all import statements

marketplace/commercetools-products/src/product-card.js
```diff
@@ -0,0 +1,231 @@
...
+  let val = nameAllLocales.find(item => item.locale === locale);
+
+  if (val === undefined) {
+    val = nameAllLocales[0];
+  }
+
+  return val.value;
+}
+
+function ProductCard({ sku, locale, removeItem, index, sortable, isDisabled, projectKey }) {
```
giotiskl: again here potentially we could split the components into files to avoid having a file over 200+ lines

marketplace/commercetools-products/extension.json
```diff
@@ -0,0 +1,51 @@
...
+      },
+      {
+        "id": "clientId",
+        "name": "commercetools Authentication Client ID",
+        "type": "Symbol",
+        "required": true
+      },
+      {
+        "id": "clientSecret",
+        "name": "commercetools Authentication Client Secret",
```
jelz: Is it safe to have this value exposed in the frontend?

marketplace/commercetools-products/src/apollo.js
```diff
@@ -0,0 +1,41 @@
+import SdkAuth, { TokenProvider } from '@commercetools/sdk-auth';
+import { ApolloClient } from 'apollo-client';
+import { InMemoryCache } from 'apollo-cache-inmemory';
+import { createHttpLink } from 'apollo-link-http';
+import { setContext } from 'apollo-link-context';
+
+export async function getApolloClient({ authUri, projectKey, clientId, apiUri, clientSecret }) {
+  // @todo how do I get the refreshed token?
```
jelz: Can you elaborate? Is it possible for a long session to expire a token? Should we poll for its validity?

marketplace/commercetools-products/src/ct-product-query.graphql
```diff
@@ -0,0 +1,27 @@
+query productsList($skus: [String!]) {
```
jelz: Is this the only query? Can we use the REST API instead? I don't see too much value in bundling Apollo if we can get the same data with a single `fetch` call.

marketplace/commercetools-products/src/dialog.js
```diff
@@ -0,0 +1,91 @@
...
+import { Button } from '@contentful/forma-36-react-components';
+
+export class CommerceToolsDialog extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.pickerContainerRef = React.createRef();
+    this.picker = null;
+
+    this.onCancel = this.onCancel.bind(this);
```
jelz: I think you should be able to define these methods as properties ( `onCancel = () => { }` in the class definition) so you don't need to bind.

marketplace/commercetools-products/src/dialog.js
```diff
@@ -0,0 +1,91 @@
...
+      </React.Fragment>
+    );
+  }
+
+  onCancel() {
+    this.props.extension.close(null);
+  }
+
+  onInsert() {
+    this.picker.getSelectedItems().then(items => {
```
jelz: any reason for not using `async/await` here and below?

marketplace/commercetools-products/src/dialog.js
```diff
@@ -0,0 +1,91 @@
...
+    this.onInsert = this.onInsert.bind(this);
+  }
+
+  render() {
+    return (
+      <React.Fragment>
+        <div ref={this.pickerContainerRef} className="picker-container" />
+
+        <div className="picker-controls">
+          <Button onClick={this.onInsert} buttonType="positive" style={{ marginRight: 16 }}>
```
jelz: Avoid styles definitions as inline object literals. Extract them to constant

```
const styles = {
  pickerButton: { marginRight: 16 },
  somethingElse: { color: 'red' }
}
```

Potentially in a separate file.

Consider using https://github.com/emotion-js/emotion if you see a benefit (utility vs bundle size addition).

marketplace/commercetools-products/src/field.js
```diff
@@ -0,0 +1,203 @@
...
+  locale,
+  sortable,
+  isDisabled,
+  projectKey
+}) {
+  return (
+    <ol className="product-list">
+      {items.map((sku, index) => {
+        return (
+          <li key={index} className="product-list__item">
```
jelz: Use something more stable (SKU?)

jelz: You're using CSS class here. I'd either stick to CSS everywhere or convert to CSS-in-JS everywhere.

marketplace/commercetools-products/src/field.js
```diff
@@ -0,0 +1,203 @@
...
+      this.unsubscribeOnIsDisabled();
+    }
+  }
+
+  onDialogOpen() {
+    const { isSingle, extension } = this.props;
+
+    extension.dialogs
+      .openExtension({
+        id: 'commercetools-products', // @todo Check with the team how to make this work with Marketplace installs
```
jelz: Stable ID is not guaranteed for apps. Use `sdk.ids.extension` to get in on runtime.

marketplace/commercetools-products/src/dialog.js
```diff
@@ -0,0 +1,91 @@
...
+    this.pickerContainerRef = React.createRef();
+    this.picker = null;
+
+    this.onCancel = this.onCancel.bind(this);
+    this.onInsert = this.onInsert.bind(this);
+  }
+
+  render() {
+    return (
+      <React.Fragment>
```
jelz: You could use `<> </>` shortcurt here and in the following files.

marketplace/commercetools-products/src/index.html
jelz: This looks shady: `herokuapp`, no version in the filename (can it break w/o us knowing?)

realityking: This is the product picker provided by Commerce Tools. Since they wrote this because of this extension (though they'll use it for other integrations in the future) it's not deployed nicely (yet). I'll raise this with them before we publish it to the Marketplace.

marketplace/commercetools-products/src/index.js
realityking: Done. But if this is the style we're going for, we should enforce it in `@contentful/exlint-config-extension` using https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md

marketplace/commercetools-products/extension.json
realityking: Yes, it can be scoped to only allow reading products.

marketplace/commercetools-products/src/field.js
realityking: Works like a charm 👍

marketplace/commercetools-products/src/dialog.js
realityking: This is one of those things I prefer not to do without a linter enforcing it. Keeping these rules in my head (when I only rarely work in repos with this code style) is a lot. If it's a rule in `@contentful/eslint-config-extension`, that'd be cool.

jelz:
Closing this one

PR #190, opened 2019-08-23
DAM apps smooth D&D

jelz:
This makes DAM apps drag and drop really smooth, grid-base and also fixes very glitchy/blinkly firefox

### DIFF
--- a/marketplace/shared-dam-app/src/Editor/SortableComponent.tsx
+++ b/marketplace/shared-dam-app/src/Editor/SortableComponent.tsx
@@ -3,7 +3,6 @@ import { SortableContainer, SortableElement, SortableHandle } from 'react-sortab
 import { css } from 'emotion';
 import arrayMove from 'array-move';
 import { IconButton, Card } from '@contentful/forma-36-react-components';
-import tokens from '@contentful/forma-36-tokens';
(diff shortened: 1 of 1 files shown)


marketplace/shared-dam-app/src/Editor/SortableComponent.tsx
```diff
@@ -83,21 +83,47 @@ const SortableItem = SortableElement<SortableElementProps>((props: SortableEleme
...
+      const [url, alt] = props.makeThumbnail(resource, props.config);
+      const item = { url, alt, key: `url-unknown-${index}` };
+      const counts = { ...acc.counts };
+
+      // URLs are used as keys.
+      // It is possible to include the same image more than once.
+      // We count usages of the same URL and use the count in keys.
+      // This can be considered an edge-case but still - should be covered.
+      if (url) {
+        counts[url] = counts[url] || 1;
```
davidfateh: Should the default be `0` here instead of `1`? You are incrementing on line 100.

jelz: doesn't really matter, I guess my keys are 1-indexed ;)

PR #195, opened 2019-09-02
Improve typeform extension to be ready to be published to the Marketplace

realityking:
Changes made

- Created `MARKETPLACE.md` for the listing
- Renamed to `typeform-form` to be more in line with other extensions
- updated all dependencies
- changed to store the public URL of the form, not its ID. This makes consuming the content much easier as the ID doesn't have to be resolved through the API.
- support disabled state
- support `onValueChanged`
- fixed handling of API errors


marketplace/typeform-form/src/app.js
```diff
@@ -13,26 +13,41 @@ export class App extends React.Component {
...
   }
 
   componentDidMount() {
-    const { accessToken, workspaceId } = this.props.parameters;
+    const { sdk, parameters } = this.props;
+    const { accessToken, workspaceId } = parameters;
     const { value } = this.state;
 
+    // Handle external changes (e.g. when multiple authors are working on the same entry).
+    sdk.field.onValueChanged(value => {
```
suevalov:
```
componentDidMount() {
   this.unsubscribeOnValue = dk.field.onValueChanged();
}

componentWillUnmount() {
   if (this.unsubscribeOnValue) {
      this.unsubscribeOnValue();
   }
}

```

marketplace/typeform-form/src/app.js
```diff
@@ -13,26 +13,41 @@ export class App extends React.Component {
...
+      this.setState({ value: value });
+    });
+
+    // Disable editing (e.g. when field is not editable due to R&P).
+    sdk.field.onIsDisabledChanged(isDisabled => {
```
suevalov: the same as above

suevalov:
Was the extension published already? If so and somebody uses it then I would argue against renaming.

realityking:
Huh, TIL about being able to unsubscribe from these. Thanks!

The extension hasn't been published. Unless someone figured out the URL and installed it manually. I think it's safe to change.

PR #196, opened 2019-09-02
Add some basic integration tests

realityking:
These tests validate some basics assumptions about the source and destination file structure.

I originally also had a test to check for `README.md` and `MARKETPLACE.md` but this is no longer required for apps so I'm leaving that one out.

Most interesting is probably that I've moved `shared-dam-app` to a different top level folder `lib`. This nicely separates shared libraries from actual extensions/apps.


suevalov:
Since we're preparing a release and working on these files, I'd prefer to put in PR on hold. ~ 1 week or so. Does it work for you @realityking ?

jelz:
Agree w/ @suevalov ; we would also have to QA again before the release

realityking:
Makes sense. Didn't quite have it on my radar how this affects the release. 

I'll back out the commits touching the `shared-dam-app` and the tests that need the change to get the infrastructure

suevalov:
@realityking can you please rebase the branch so we can take proceed with a review?

realityking:
@suevalov Rebased. Sorry for the delay.

PR #211, opened 2019-09-26
Image metadata - app configuration

giotiskl:


samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,156 @@
...
+import { styles } from './styles';
+
+export class AppView extends Component {
+  static propTypes = {
+    sdk: PropTypes.object.isRequired
+  };
+
+  state = {
+    allContentTypesIds: [],
+    appIsInstalled: null,
```
davidfateh: Why is the default not `false`?

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,156 @@
...
+    const appIsInstalled = await app.isInstalled();
+
+    const allContentTypes = await sdk.space.getContentTypes();
+    const allContentTypesIds = allContentTypes.items.map(({ sys: { id } }) => id);
+
+    // Following eslint error is caused due to using async/await
+    // eslint-disable-next-line react/no-did-mount-set-state
+    this.setState({ appIsInstalled, allContentTypesIds });
+
+    if (!appIsInstalled) {
```
davidfateh: Remove this `if` statement but keep the inner code. You want to give `onConfigure` a callback regardless of if the app is installed or not. This is because when I hit save (after the app is installed), it will still call this method.

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,156 @@
...
+
+    if (!contentTypeName) {
+      this.props.sdk.notifier.error('Provide a name for the content type.');
+      return false;
+    }
+
+    const isContentTypeIdTaken = allContentTypesIds.includes(contentTypeId);
+    if (isContentTypeIdTaken) {
+      this.props.sdk.notifier.error(
+        `Id "${contentTypeId}" is taken. Try a different name for the content type`
```
davidfateh:
```suggestion
        `ID "${contentTypeId}" is taken. Try a different name for the content type.`
```

davidfateh: `Id` looks like a typo of `I'd`. Using all caps differentiates it means ID.

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,156 @@
...
+      );
+      return false;
+    }
+
+    const { sdk } = this.props;
+    const contentType = await sdk.space
```
davidfateh: From my comment on line 36, you will need to alter this logic to instead do this work only if the content type has been removed.

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,156 @@
...
+            type: 'Object'
+          }
+        ]
+      })
+      .catch(() =>
+        this.props.sdk.notifier.error(`Failed to create content type "${contentTypeName}"`)
+      );
+
+    // Set the newly created content type's state to "Published"
+    await sdk.space
```
davidfateh: Same comment as line 58.

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,156 @@
...
+      .updateContentType(contentType)
+      .catch(() =>
+        this.props.sdk.notifier.error(`Failed to publish content type "${contentTypeName}"`)
+      );
+
+    this.setState({ appIsInstalled: true });
```
davidfateh: This might not be correct. There is still a possibility of failure. We may need to revisit how this works.

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,156 @@
...
+            ]
+          }
+        }
+      }
+    };
+  };
+
+  setContentTypeName = ({ target: { value } }) =>
+    this.setState({
+      contentTypeId: camelCase(value),
```
davidfateh: Just a thought: Instead of using `camelCase` what if we showed a list of content types that a user could select from and then pass the internal ID when the user selects. This way the user never has to manually input the ID of the content type and we can minimize user error in this case.

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,156 @@
...
+    });
+
+  render() {
+    const { appIsInstalled, allContentTypesIds, contentTypeId, contentTypeName } = this.state;
+
+    return (
+      <>
+        <div className={styles.background} />
+        <div className={styles.body}>
+          <div>
```
davidfateh: I believe this `div` can be removed and we could just use `Typography`.

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,156 @@
...
+                  contentTypeId={contentTypeId}
+                  contentTypeName={contentTypeName}
+                  setContentTypeName={this.setContentTypeName}
+                />
+              )}
+            </Typography>
+          </div>
+        </div>
+        <div className={styles.logo}>
+          <img src="https://image.flaticon.com/icons/svg/189/189089.svg" alt="logo" />
```
davidfateh: 1. We should check that we can actually use this image.

2. We should include this file right in the package so we don't have it disappear on us in the future.

samples/image-metadata/src/components/AppView/ConfigurationContent.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { Heading, Paragraph } from '@contentful/forma-36-react-components';
+import { Divider } from '../Divider';
+import { styles } from './styles';
+
+export const ConfigurationContent = () => (
+  <>
+    <Heading className={styles.heading}>Getting started</Heading>
+    <Paragraph>
+      The Image Wrapper app is now installed. We have set up a demo content type for you with a
```
davidfateh: I suggest saying this instead because a user could come back to this page later to change a config.

```suggestion
      The Image Wrapper app is installed. We have set up a demo content type for you with a
```
Also, I feel like we shouldn't use the word 'demo'. We are installing a content type that is certainly usable for non-demo purposes.

samples/image-metadata/src/components/AppView/ConfigurationContent.js
```diff
@@ -0,0 +1,32 @@
...
+      data).
+    </Paragraph>
+    <Paragraph className={styles.paragraph}>
+      The metadata field is a plain Object type field with its appearance set to be that of the
+      current app and the image field id parameter pointing to the image field, as the data it holds
+      is connected to that image.
+    </Paragraph>
+    <Paragraph className={styles.paragraph}>
+      To create a new image with focal point data, simply create a new entry for the content type we
+      set up for you in the Content tab. Then create a Reference field in your other Content Model
```
davidfateh:
```suggestion
      set up for you on the Content page. Then create a Reference field in your other Content Model
```

I think the word 'tab' is not clear.

samples/image-metadata/src/components/AppView/ConfigurationContent.js
```diff
@@ -0,0 +1,32 @@
...
+      that needs to reference this image and make it point to the entry you just created.
+    </Paragraph>
+    <Divider />
+    <Heading className={styles.heading}>Uninstallation &amp; Cleanup</Heading>
+    <Paragraph>
+      Please note that if you decide to uninstall this app you will have to take care of deleting
+      the demo content type we set up for you during the installation process. This might require
+      you to delete any entries for that content type you have created first.
```
davidfateh: It feels like we don't really need this paragraph. This is information that will actually be populated in the uninstall modal.

samples/image-metadata/src/components/AppView/InstallationContent.js
```diff
@@ -0,0 +1,47 @@
...
+import { styles } from './styles';
+
+export function InstallationContent({
+  allContentTypesIds,
+  contentTypeId,
+  contentTypeName,
+  setContentTypeName
+}) {
+  const validationMessage = allContentTypesIds.includes(contentTypeId)
+    ? `A content type with id "${contentTypeId}" already exists. Try a different name for the demo content type.`
```
davidfateh:
```suggestion
    ? `A content type with ID "${contentTypeId}" already exists. Try a different name for the demo content type.`
```

Also, another instance of the word 'demo'.

samples/image-metadata/src/components/AppView/ConfigurationContent.js
```diff
@@ -0,0 +1,32 @@
...
+    </Paragraph>
+    <Paragraph className={styles.paragraph}>
+      To create a new image with focal point data, simply create a new entry for the content type we
+      set up for you in the Content tab. Then create a Reference field in your other Content Model
+      that needs to reference this image and make it point to the entry you just created.
+    </Paragraph>
+    <Divider />
+    <Heading className={styles.heading}>Uninstallation &amp; Cleanup</Heading>
+    <Paragraph>
+      Please note that if you decide to uninstall this app you will have to take care of deleting
```
davidfateh:
```suggestion
      Please note that if you decide to uninstall this app, you will have to take care of deleting
```

samples/image-metadata/src/components/FocalPointDialog/FocalPointDialog.js
```diff
@@ -55,6 +54,11 @@ export class FocalPointDialog extends Component {
   };
 
   onImageClick = e => {
+    const imageWasClicked = this.imgRef.current === e.target;
```
davidfateh: Is `this.imgReg` always going to exist? Should we check it's there first before trying to access `current`?

davidfateh:
Overall, I like the approach. I do think though we should move away from telling users this is a demo content type and just tell them to use this content type. I think that will help in the future because if we want to make changes/updates to this app, we can do so knowing users are using the content type that was installed. If we say they can create their own type, then it makes it a little trickier.

samples/image-metadata/src/components/AppView/AppView.js
lehnerm: Why not just check inside the callback if the app was already installed, and in that case not provision the CT?

samples/image-metadata/src/components/AppView/AppView.js
giotiskl: This is a placeholder :) Malin is making a real logo it'll be removed

samples/image-metadata/src/components/AppView/AppView.js
lehnerm: To make this consistent, we would need to have a callback along the lines of `onConfigurationCompleted()`. All other solutions i can come up with face trouble with race conditions in the backend.

samples/image-metadata/src/components/AppView/InstallationContent.js
giotiskl: Done

samples/image-metadata/src/components/AppView/AppView.js
giotiskl: Done. (With the discussed hack).

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,156 @@
...
+          {
+            id: 'imageMetadata',
+            name: 'Image metadata',
+            required: true,
+            type: 'Object'
+          }
+        ]
+      })
+      .catch(() =>
+        this.props.sdk.notifier.error(`Failed to create content type "${contentTypeName}"`)
```
lehnerm: This needs to cause a `return false` to indicate the configuration was unsuccessful

samples/image-metadata/src/components/AppView/AppView.js
giotiskl: Done.

samples/image-metadata/src/components/AppView/AppView.js
giotiskl: Changed based on the discussed hack

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,179 @@
...
+      );
+
+    // TODO: hack that determines when the app has been successfully installed.
+    // To be done away with once the post installation hook is implemented.
+    const appIsInstalledInterval = setInterval(async () => {
+      const appIsInstalled = await sdk.platformAlpha.app.isInstalled();
+
+      if (
+        appIsInstalled &&
+        this.state.appInstallationState === APP_INSTALLATION_STATE.NOT_INSTALLED
```
davidfateh: Is there a chance that the app gets stuck in a state where it is installed but for some reason this condition is false meaning that the `interval` just keeps running forever? It seems like it can't be the case...

samples/image-metadata/src/components/AppView/AppView.js
giotiskl: Done

samples/image-metadata/src/components/AppView/AppView.js
giotiskl: Nope, cannot happen.

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,179 @@
...
+      })
+      .catch(() =>
+        this.props.sdk.notifier.error(`Failed to create content type "${contentTypeName}"`)
+      );
+
+    // Set the newly created content type's state to "Published"
+    await sdk.space
+      .updateContentType(contentType)
+      .catch(() =>
+        this.props.sdk.notifier.error(`Failed to publish content type "${contentTypeName}"`)
```
davidfateh: @lehnerm 's comment above on line 100 is also relevant to here as well. Should return `false`.

samples/image-metadata/src/components/AppView/AppView.js
davidfateh: I misunderstood what was going on before. I think this comment is now irrelevant.

samples/image-metadata/src/components/AppView/AppView.js
giotiskl: Done.

samples/image-metadata/src/components/AppView/AppView.js
giotiskl: Done.

samples/image-metadata/src/components/AppView/AppView.js
jelz: In addition to that I'd also call `onConfigure` as soon as possible, not after async actions

samples/image-metadata/src/components/AppView/AppView.js
```diff
@@ -0,0 +1,184 @@
...
+
+    if (!contentTypeName) {
+      this.props.sdk.notifier.error('Provide a name for the content type.');
+      return false;
+    }
+
+    const isContentTypeIdTaken = allContentTypesIds.includes(contentTypeId);
+    if (isContentTypeIdTaken) {
+      this.props.sdk.notifier.error(
+        `ID "${contentTypeId}" is taken. Try a different name for the content type`
```
jelz: ACtually this reminds me that we could also allow users to customize the ID (API name). IDs are used in CDN queries, SDKs etc. and some people might want to use some ID that follows their content type ID convention (instead of generated one). I'd render two inputs, one for name and one for ID, with ID automatically generated (`camelCase(name)` is fine) until you touch the ID field.

giotiskl: Done

giotiskl:
> I'd rename "Image Wrapper" to "Focal Point Wrapper" or something like that. "Image Wrapper" is waay too generic.

Done and wrote tests

PR #215, opened 2019-10-11
Call setReady in Netlify

giotiskl:


marketplace/netlify-build/src/app/index.js
```diff
@@ -64,14 +64,17 @@ export default class NetlifyAppConfig extends React.Component {
...
-      ready: true,
-      config,
-      enabledContentTypes,
-      contentTypes: contentTypesResponse.items.map(ct => [ct.sys.id, ct.name]),
-      netlifySites: uniqBy(netlifySites, s => s.id),
-      ticketId
-    });
+    this.setState(
+      {
+        ready: true,
```
jelz: Remove this and remove all code related to it

marketplace/netlify-build/src/app/index.js
```diff
@@ -64,14 +64,17 @@ export default class NetlifyAppConfig extends React.Component {
...
+        config,
+        enabledContentTypes,
+        contentTypes: contentTypesResponse.items.map(ct => [ct.sys.id, ct.name]),
+        netlifySites: uniqBy(netlifySites, s => s.id),
+        ticketId
+      },
+      app.setReady
```
jelz: I'd wrap with a function

marketplace/netlify-build/src/app/index.js
giotiskl: done

marketplace/netlify-build/src/app/index.js
giotiskl: any reason to go for the more verbose variation?

jelz: so we make sure it's called with 0 args (just to be safe)

marketplace/netlify-build/src/app/index.js
```diff
@@ -171,39 +177,33 @@ export default class NetlifyAppConfig extends React.Component {
 
   render() {
     const disabled = !this.state.token;
+    console.log('config', this.state.config);
```
jelz: delete

PR #216, opened 2019-10-11
Call setReady for DAM Apps

giotiskl:


lib/shared-dam-app/src/AppConfig/AppConfig.tsx
```diff
@@ -121,13 +121,16 @@ export default class AppConfig extends React.Component<Props, State> {
...
-    });
+    this.setState(
+      {
+        ready: true,
+        contentTypes: filteredContentTypes,
+        compatibleFields,
+        selectedFields: currentStateToSelectedFields(currentState || {}),
+        parameters: toInputParameters(this.props.parameterDefinitions, parameters)
+      },
+      (platformAlpha.app as any).setReady
```
suevalov: I would leave a todo comment.

lib/shared-dam-app/src/AppConfig/AppConfig.tsx
```diff
@@ -121,13 +121,16 @@ export default class AppConfig extends React.Component<Props, State> {
...
-    this.setState({
-      ready: true,
-      contentTypes: filteredContentTypes,
-      compatibleFields,
-      selectedFields: currentStateToSelectedFields(currentState || {}),
-      parameters: toInputParameters(this.props.parameterDefinitions, parameters)
-    });
+    this.setState(
+      {
+        ready: true,
```
jelz: This state property and any code related to it should be gone.

lib/shared-dam-app/src/AppConfig/AppConfig.tsx
jelz: I would wrap it with a function still instead of passing a function straight out of the SDK

PR #217, opened 2019-10-11
Call setReady in Optimizely

giotiskl:


marketplace/optimizely/src/AppPage/index.js
```diff
@@ -80,15 +80,18 @@ export default class AppPage extends React.Component {
...
-      config: {
-        contentTypes: enabledContentTypes,
-        optimizelyProjectId: currentParameters
-          ? currentParameters.optimizelyProjectId
-          : prevState.optimizelyProjectId
-      }
-    }));
+    this.setState(
+      prevState => ({
+        allContentTypes,
```
jelz: Do we have a loading state for these?

giotiskl: I didn't find any sort of custom loading logic for optimizely, it simply differentiates on which component to render based on whether an optimizely client has been provided, but that piece of logic is still needed

PR #222, opened 2019-10-26
Improved single selection view

acasazza:
Hi there! I'd fix the white spaces about my mistake and I improved the single view when the sku is selected.

### Before
Screen Shot 2019-10-26 at 11 32 00

### After
Screen Shot 2019-10-26 at 11 21 09


acasazza:
Hi @davidfateh, Done! 😉 Thanks!

PR #223, opened 2019-10-30
Remove instance param from focal point app

giotiskl:


marketplace/image-focal-point/src/index.js
```diff
@@ -122,10 +119,9 @@ export class App extends React.Component {
 
   render() {
     const { sdk } = this.props;
-    const { imageFieldId } = sdk.parameters.instance;
-    const imageField = getField(sdk.contentType, imageFieldId);
+    const imageField = getField(sdk.contentType, IMAGE_FIELD_ID);
     const isImageField = isCompatibleImageField(imageField);
-    const hasValidConfig = !!imageFieldId && isImageField;
+    const hasValidConfig = isImageField;
```
jelz: I'd just use `isImageField` and drop this line.

PR #232, opened 2019-10-30
Validate and save upon field change

philmirez:
Validate and save upon field change as opposed to extension load.
This was disrupting my editorial team's workflow. 
After updating the schema, once the extension was in view then the entry status would change from publish to pending changes.
This proposed file change will only change the entry status upon a field being changed rather than the extension being loaded.


davidfateh:
@philmirez you should be all set. Since you probably installed this using `srcdoc` you'll want to update that source inside the Contentful web app.

PR #245, opened 2019-11-08
Default token option for Frontify app

jelz:
In Frontify it sometimes makes sense to use separate keys per field to access different collections.

This PR adds the default option in app configuration that can be then overridden on the field level.

@brunschgi could you also please take a look?


brunschgi:
Hi @jelz
Thanks a lot for the change – I just tried your change on our instance and it works like a charm!
I also pushed the changes to our fork.
Best, Remo

PR #256, opened 2019-11-08
Add Dropbox app

jelz:


lib/shared-dam-app/src/AppConfig/AppConfig.tsx
```diff
@@ -178,42 +178,46 @@ export default class AppConfig extends React.Component<Props, State> {
...
+          </Typography>
+        )}
         <Typography>
           <Heading>Assign to fields</Heading>
           {contentTypes.length > 0 ? (
             <Paragraph>
-              This app can only be used with <strong>JSON object</strong> fields. Select which ones
-              you’d like to enable this app.
+              This app can only be used with <strong>JSON object</strong> fields. Select for which
+              ones you’d like to enable this app.
```
davidfateh: "Select which JSON fields you'd like to enable for this app."

or 

"Select which ones you'd like to enable for this app."

I like the first one the best because it's specific but I would definitely move the word 'for'.

marketplace/dropbox-assets/src/index.html
```diff
@@ -0,0 +1,24 @@
...
+        font-size: 100%;
+        font: inherit;
+        vertical-align: baseline;
+        font-family: system, sans-serif;
+      }
+    </style>
+  </head>
+  <body>
+    <div id="root"></div>
+    <script src="https://www.dropbox.com/static/api/2/dropins.js" id="dropboxjs" data-app-key="g1i5ql4pfnwohig"></script>
```
davidfateh: Any specific reason to use the script instead of importing the npm package?

jelz: Just following the instruction: https://www.dropbox.com/developers/chooser

There is no official package for Chooser. Official SDK doesn't support it (https://www.npmjs.com/package/dropbox). I guess we're better off with Dropbox CDN.

davidfateh: Gotcha, makes sense.

PR #260, opened 2019-11-22
Default value extension next version

jelz:


floelhoeffel:
Tested it from the branch, works

PR #264, opened 2019-12-06
Implement Commerce Layer app

giotiskl:


marketplace/commercelayer-products/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License
+
+Copyright (c) 2010-2019 Michel Bitter, Bleeve B.V.
```
davidfateh: Why is this also Michel? I saw this name in the Commercetools app as well. We should figure out exactly which one his name belongs to.

marketplace/commercelayer-products/src/dataTransformer.js
```diff
@@ -0,0 +1,30 @@
...
+    sku: code,
+    externalLink: `${projectUrl}/admin/skus/${id}/edit`
+  };
+};
+
+/**
+ * Transforms the API response of CommerceLayer into
+ * the product schema expected by SkuPicker Preview components
+ */
+export const previewDataTransformer = projectUrl => ({ id, imageUrl, name, code }) => {
```
davidfateh: Why do these functions only differ slightly? It seems like they could potentially be merged.

marketplace/commercelayer-products/src/index.js
```diff
@@ -0,0 +1,162 @@
...
+ * @see fetchSKUs for a more detailed explanation.
+ */
+async function makeCommerceLayerClient({
+  parameters: {
+    installation: { apiEndpoint, clientId, clientSecret }
+  }
+}) {
+  const auth = await CLayerAuth.integration({
+    clientId,
+    clientSecret,
```
lehnerm: OAuth client secrets should not be in the frontend. Do they not support an implicit flow?

marketplace/commercelayer-products/src/index.js
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/**
+ * Fetches the product previews for the products selected by the user
+ */
+const fetchProductPreviews = async function fetchProductPreviews(skus, config) {
+  if (!skus.length) {
+    return [];
+  }
+  await makeCommerceLayerClient({ parameters: { installation: config } });
```
lehnerm: Since api installations can provide whatever they like, we should validate installation parameters before use. You also already have a method for that :)

marketplace/commercelayer-products/src/index.js
giotiskl: They do, shared-sku-picker doesn't :+1: We should probably revisit Commercetools then it's the same story there

giotiskl: **Update**: now use the `salesChannel` flow as discussed which doesn't require a client secret

marketplace/commercelayer-products/src/index.js
giotiskl: Done

marketplace/commercelayer-products/src/dataTransformer.js
giotiskl: So the problem is that when you fetch data through a direct HTTP req the returned schema format differs from when you request the same data using their JS SDK. Doing both is unavoidable due to the way the Commerce Layer API implements pagination.

I explain more on that in the comments above the `fetchSKUs` function in index.js

marketplace/commercelayer-products/LICENSE
giotiskl: Updated to Contentful. This is a leftover from Cloudinary I guess which served as the base project for both... some copy pasting leftover

marketplace/commercelayer-products/src/dataTransformer.js
giotiskl: I unified them into 1 function.

marketplace/commercelayer-products/src/index.js
lehnerm: awesome :+1:

marketplace/commercelayer-products/LICENSE
davidfateh: So I guess we should also change it in the other LICENSE.

marketplace/commercelayer-products/src/index.js
jelz: If we don't require it then why do we ask people to provide it in the configuration screen?

marketplace/commercelayer-products/src/index.js
jelz: Why we don't pass `config` directly? I don't think `makeCommerceLayerClient` needs to know about either parameters or installation parameters.

marketplace/commercelayer-products/extension.json
```diff
@@ -0,0 +1,31 @@
+{
+  "id": "commercelayer-products",
+  "name": "Commerce Layer",
+  "fieldTypes": ["Symbol"],
```
jelz: The value is not really used if we use app endpoints but I'd still do `["Symbol", "Symbols"]` if we ever decide to handle `AppDefinition` in the tooling.

marketplace/commercelayer-products/src/index.js
```diff
@@ -0,0 +1,170 @@
...
+
+  const res = await fetch(URL, {
+    headers: {
+      Accept: 'application/vnd.api+json',
+      Authorization: `Bearer ${auth.accessToken}`
+    },
+    method: 'GET'
+  });
+
+  const result = await res.json();
```
jelz: `return res.json()`

marketplace/commercelayer-products/src/index.js
```diff
@@ -0,0 +1,170 @@
...
+  if (!skus.length) {
+    return [];
+  }
+  await makeCommerceLayerClient({ parameters: { installation: config } });
+  const result = await CLayer.Sku.where({ code_in: skus.join(',') }).all();
+  return result.toArray().map(dataTransformer(config.apiEndpoint));
+};
+
+async function renderDialog(sdk) {
+  const ID = 'dialog-root';
```
jelz: Declare outside of this function as `DIALOG_ID`

marketplace/commercelayer-products/extension.json
giotiskl: done.

marketplace/commercelayer-products/src/index.js
giotiskl: done.

marketplace/commercelayer-products/src/index.js
giotiskl: done.

giotiskl:
@jelz worked on the comments of yesterday. Also did a couple of small changes to Commercetools to avoid having an extra PR for that

PR #266, opened 2019-12-10
Disable asset preview for SVGs during upload

giotiskl:


samples/image-uploader/src/components/ProgressView/index.js
```diff
@@ -1,25 +1,30 @@
...
-    "--uploadProgress": `${props.uploadProgress}%`
-  }
+    '--uploadProgress': `${props.uploadProgress}%`
+  };
 
   return (
     <div className="viewport">
-      <main className="progress-view" style={background}>
+      <main className="progress-view" style={progressViewStyles}>
+        {!isSVG && <img className="uploaded-image" src={imageUrl} alt="image being uploaded" />}
```
davidfateh: Why not just make this a ternary instead of having two different conditions?

giotiskl: voila

PR #267, opened 2019-12-12
Gatsby cleanup

davidfateh:


marketplace/gatsby-preview/src/AppConfig.js
```diff
@@ -57,7 +57,7 @@ export default class AppConfig extends React.Component {
         webhookUrl: params.webhookUrl || '',
         authToken: params.authToken || ''
       };
-    });
+    }, app.setReady);
```
giotiskl: here Jakub had argued before we should better use `() => app.setReady()` to avoid inadvertently passing down any arguments just in case

marketplace/gatsby-preview/src/AppConfig.js
```diff
@@ -47,17 +47,20 @@ export default class AppConfig extends React.Component {
...
-        }, prevState.checkedContentTypes),
-        previewUrl: params.previewUrl || '',
-        webhookUrl: params.webhookUrl || '',
-        authToken: params.authToken || ''
-      };
-    });
+    this.setState(
+      prevState => {
+        return {
+          checkedContentTypes: items.reduce((acc, ct) => {
```
jelz: Is `acc` an array or an object? Something is off here.

davidfateh: `acc` is an object in this sense which looks something like `{name: string, checked: boolean}` and is built off the previous state, `prevState.checkedContentTypes`. I actually didn't change any of this logic here, I only added `() => app.setReady()`. Prettier took over the rest...

jelz: Ah true, I misunderstood the code.

I'd still do:

```
return {
  ...acc,
  [ct.sys.id]: { name: ct.name, checked: previouslyCheckedTypes.includes(ct.sys.id)
}
```

so we don't mutate `acc`

davidfateh: Changed here: https://github.com/contentful/extensions/pull/267/commits/b36fb56f8ac178c078978726487fb8d2be86b3b7

PR #269, opened 2019-12-17
Integrate CommerceTools developed product picker

giotiskl:
The CommerceTools has been updated to use the product picker developed my CommerceTools (as opposed to using our shared-sku-picker widget).

@davidfateh as you might have to do some work on this during the holidays, feel free to reach out to @floelhoeffel. The current state is as follows:
* Product picker is rendered
* User selects a product
* Only SKU gets persisted.

Things to consider are:
* how do we deal with rendering the categories picker as opposed to the product picker?
* what's the deal with multiple product selection, is that an option anymore?


marketplace/commercetools-sku/src/index.js
```diff
@@ -14,6 +14,34 @@ function makeCTA(fieldType) {
...
+  if (parameters.clientId.length < 1) {
+    return 'Provide your Commercetools client ID.';
+  }
+
+  if (parameters.clientSecret.length < 1) {
+    return 'Provide your Commercetools client secret.';
+  }
+
+  if (parameters.apiEndpoint.length < 1) {
+    return 'Provide the Commercetools API endpoint.';
```
davidfateh: The messages above say "your". Is this endpoint and the one below going to be static for all user? If so, can't we just hardcode it in?

marketplace/commercetools-sku/src/index.js
```diff
@@ -73,42 +99,36 @@ async function renderDialog(sdk) {
...
+  };
+
+  const ctPicker = new window.CTPicker(pickerOptions, container);
+  try {
+    sdk.window.updateHeight(660);
+    const skus = await ctPicker.show();
+    sdk.close(skus.map(({ masterVariant: { sku } }) => sku));
+  } catch (error) {
+    if (error !== 'cancel') {
+      // Widget is going to throw if the user closes the product picking dialog
```
davidfateh: Strange.

marketplace/commercetools-sku/src/index.html
```diff
@@ -23,6 +23,10 @@
   </head>
   <body>
     <div id="root"></div>
+    <script
+      src="https://commercetools-picker-beta.herokuapp.com/ct-picker.min.js"
```
lehnerm: this needs to be `early-access` not `beta` :rofl:

marketplace/commercetools-sku/src/index.js
giotiskl: I'm not clear on this either tbh, I just kept it because it was like this in Rouven's PR, but we should clear this up with Ivo from CT before merging.

giotiskl: @davidfateh so the update is that they can differ.. based on region (e.g. EU vs US), I'd say we can leave the fields as is for now, but I think it also makes more sense to keep "the" instead of "your" since it's not really individual?

marketplace/commercetools-sku/src/index.js
giotiskl: it is how it is :man_shrugging:

marketplace/commercetools-sku/src/index.js
davidfateh: Sounds good.

PR #272, opened 2020-01-16
Update Apps SDK

jelz:
- [x] Update SDK everywhere and use all proper methods (no `platformAlpha` anymore)
- [x] Switch to `LOCATION_APP_CONFIG`
- [x] Use `dialogs.openCurrentApp()` instead of extension dialog
- [x] Use `app.onConfigurationCompleted()` in focal point
- [x] Remove use of `app.getCurrentState()` and replace it with custom logic on top of `space.getEditorInterfaces()`; the method will be removed in the next minor version bump


marketplace/gatsby-preview/src/AppConfig.js
```diff
@@ -14,60 +15,70 @@ import {
 import GatsbyIcon from './GatsbyIcon';
 import styles from './styles';
 
+function editorInterfacesToEnabledContentTypes(eis, appId) {
+  const findAppWidget = item => item.widgetNamespace === 'app' && item.widgetId === appId;
+
+  return eis
+    .filter(ei => !!get(ei, ['sidebar'], []).find(findAppWidget))
+    .map(ei => get(ei, ['sys', 'contentType', 'sys', 'id']))
+    .filter(ctId => typeof ctId === 'string' && ctId.length > 0);
```
davidfateh: I'm seeing some similar logic as in the other apps. Would it be worth abstracting this into an `sdk.app` method for convenience to the end-user?

marketplace/image-focal-point/src/components/AppView/AppView.js
```diff
@@ -10,129 +10,109 @@ import { styles } from './styles';
...
+    });
   }
 
   installApp = async () => {
-    const { allContentTypesIds, contentTypeId, contentTypeName } = this.state;
+    const { sdk } = this.props;
+    const { isInstalled, allContentTypesIds, contentTypeId, contentTypeName } = this.state;
+
+    if (isInstalled) {
+      sdk.notifier.error('Nothing to do: the app is already fully configured.');
```
davidfateh: What about just removing this message since the user doesn't need to know that nothing happened. 

Or change it from error to the green message block since there is no error.

marketplace/netlify-build/src/app/target-state.js
```diff
@@ -1,10 +1,12 @@
...
-  const eiState = get(currentState, ['EditorInterface'], {});
-  const ctIds = Object.keys(eiState);
+export function editorInterfacesToEnabledContentTypes(eis, appId) {
+  const findAppWidget = item => item.widgetNamespace === 'app' && item.widgetId === appId;
 
-  return ctIds.filter(id => !!get(eiState, [id, 'sidebar']));
+  return eis
+    .filter(ei => !!get(ei, ['sidebar'], []).find(findAppWidget))
+    .map(ei => get(ei, ['sys', 'contentType', 'sys', 'id']))
+    .filter(ctId => typeof ctId === 'string' && ctId.length > 0);
```
davidfateh: Same comment as above.

marketplace/gatsby-preview/src/AppConfig.js
jelz: @davidfateh Yes, absolutely, I'm fully aware of the repetition. This is a part of a bigger problem: we still use `targetState` in all the apps (which is not a part of the final interface).

I don't want to introduce these "convenience" methods to the SDK since they are usecase-specific and I don't want to grow the SDK any further.

BUT the idea would be to introduce a new package (think `shared-assignment-helpers`) that would allow:

1. read `EditorInterface`s (any any other assignment entities we'll add in the future) and convert them to a pleasant data structure (for UI rendering)
2. save changes introduced in the data structure (1) back to assignment entities (so the things that is done w/ `targetState` now)
3. potentially it could export a set of components for rendering assignment UIs, but for Jan/Feb it would be a strech

That being said there's some work to do and I want to tackle it separately.

marketplace/image-focal-point/src/components/AppView/AppView.js
jelz: True, will make it green

PR #273, opened 2020-01-21
Shared SKU fixes

giotiskl:
**Done**
---
* Shopify now relies on internal variant IDs instead of SKUs
* Client secret removed from Commerce Layer
* CSS overflow problems fixed for all 3 ecommerce vendor apps.

### DIFF
--- a/marketplace/commercelayer-products/src/index.js
+++ b/marketplace/commercelayer-products/src/index.js
@@ -11,6 +11,8 @@ import { dataTransformer } from './dataTransformer';
 const DIALOG_ID = 'root';
 const PER_PAGE = 20;
 
+let accessToken = null;
+
 function makeCTA(fieldType) {
   return fieldType === 'Array' ? 'Select products' : 'Select a product';
 }
@@ -25,6 +27,22 @@ function validateParameters(parameters) {
   }
 
   return null;
+}
+
+async function getAccessToken(clientId, endpoint) {
+  if (!accessToken) {
+    /* eslint-disable-next-line require-atomic-updates */
+    accessToken = (await CLayerAuth.getIntegrationToken({
+      clientId,
+      endpoint,
+      // The empty client secret is needed for legacy reasons, as the
+      // CLayerAuth SDK will throw if not present. By setting to empty
+      // string we prevent the SDK exception and the value is ignored
+      // by the Commerce Layer Auth API.
+      clientSecret: ''
+    })).accessToken;
+  }
+  return accessToken;
 }
 
 /**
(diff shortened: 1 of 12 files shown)


marketplace/commercelayer-products/src/index.js
```diff
@@ -74,10 +41,10 @@ async function fetchSKUs(installationParams, search, pagination) {
     throw new Error(validationError);
   }
 
-  const { clientId, apiEndpoint, clientSecret } = installationParams;
-  const auth = await CLayerAuth.integration({
+  const { clientId, apiEndpoint } = installationParams;
+  const { accessToken } = await CLayerAuth.getIntegrationToken({
```
jelz: This happens twice. Can you:

- extract to a function
- add a comment mentioning why we pass an empty string, mentioning fixed version of the SDK

marketplace/shopify-sku/src/productResolvers.js
```diff
@@ -29,15 +29,25 @@ export const fetchProductPreviews = async (skus, config) => {
...
+  const res = await window.fetch(`https://${config.apiEndpoint}/api/2019-10/graphql`, {
+    method: 'POST',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+      'x-shopify-storefront-access-token': config.storefrontAccessToken
+    },
+    body: JSON.stringify({
+      query: `{nodes (ids: [${skus
+        .map(sku => `"${sku}"`)
```
jelz: Compute the string w/ IDs outside of this string template

marketplace/shopify-sku/src/productResolvers.js
```diff
@@ -29,15 +29,25 @@ export const fetchProductPreviews = async (skus, config) => {
...
+        .join(
+          ','
+        )}]) { id, __typename ...on ProductVariant { sku, image { src: originalSrc }, title, product { id, title } } }}`
```
jelz: how about some line breaks?

marketplace/commercelayer-products/src/index.js
giotiskl: Done, actually on top of that I make sure we fetch only one token, no need to fetch a new token for every req we issue

marketplace/shopify-sku/src/productResolvers.js
giotiskl: Done.

marketplace/shopify-sku/src/productResolvers.js
giotiskl: done.

giotiskl:
> @giotiskl package-lock.json change seems a little bit too big. Can you do lerna bootstrap and then `--save-exact` for the new SDK version

Did it, no change

PR #277, opened 2020-01-29
fixing setState linting error

davidfateh:
Lint doesn't recognize the `await` keyword and things this is a sync setState in mount.


dylanjha:
Hey thanks for catching that @davidfateh, not sure why I wasn't getting that error when I ran the linter 🤷‍♂ 

So I see this got merged (and I think deployed...?) about 16 hours ago. But when I log into Contentful I'm not seeing the UI updates that I created. I would expect to see the "Delete asset" button. Is there another deploy step that needs to be done?

davidfateh:
It may be because of how you have the extension installed into the space. If you are using `src` then I would check that code that URL points to got the update. If you are using `srcHost` then you will need to update the code manually in your space to get the new updates.

PR #278, opened 2020-01-30
Google analytics app

taye:


marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+import { getSidebarLocations } from './utils';
+
+export default class AppConfig extends React.Component<AppConfigParams, AppConfigState> {
+  static propTypes = {
+    sdk: PropTypes.object.isRequired
+  };
+
+  state: AppConfigState = {
+    allContentTypes: {},
+    contentTypes: {}
```
davidfateh: I would add in some defaults for the other state params.
```suggestion
    contentTypes: {},
    clientId: '',
    viewId: ''
```

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+        clientId: savedParams.clientId || '',
+        viewId: savedParams.viewId || ''
+      },
+      () => sdk.app.setReady()
+    );
+
+    app.onConfigure(() => this.configureApp());
+  }
+
+  async configureApp() {
```
davidfateh: We should add validation to this function. If you return `false` in this function, the app will cancel installation. I would check that a user has given a clientId, viewId, and picked at last one CT and slug field to actually install the app.

An example of showing an error would be:

```js
if (!clientId) {
  this.props.sdk.notifier.error('You must provide a valid clientId!');

  return false;
}
```
Notice I've also included an sdk method that will pop a red banner at the bottom of the page.

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+        viewId
+      },
+      targetState: {
+        EditorInterface
+      }
+    };
+  }
+
+  handleChange(name, value) {
+    this.setState({ [name]: value });
```
davidfateh: This function seems too generic to warrant being needed in my opinion. It basically wraps `setState` with no other functionality.

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+  }
+
+  handleContentTypeChange(prevKey, key) {
+    this.setState(prevState => {
+      const contentTypes = {};
+
+      // remove contentType[prevKey] field and replace with the new contentType
+      // key while preserving key order
```
davidfateh: Not sure what you mean by this. We shouldn't rely on keys in an object to preserve any order. If we need order, we should use an array instead.

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+            [field]: value
+          }
+        }
+      };
+    });
+  }
+
+  addContentType() {
+    this.setState(prevState => ({
+      ...prevState,
```
davidfateh: Not necessary to merge `prevState` here. `setState` will do that automatically.

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+  };
+
+  state: AppConfigState = {
+    allContentTypes: {},
+    contentTypes: {}
+  };
+
+  async componentDidMount() {
+    const { sdk } = this.props;
+    const { app } = sdk.platformAlpha;
```
jelz: Use the SDK `@latest` and change all `sdk.platformAlpha.app` to `sdk.app`

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+
+    const [{ items: spaceContentTypes }, savedParams] = await Promise.all([
+      sdk.space.getContentTypes().then(),
```
jelz: `.then()` should be gone

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+      app.getParameters()
+    ]);
+
+    const sidebarLocations = await getSidebarLocations(sdk);
```
jelz: this can be handled within `Promise.all`

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+    // add an incomplete contentType entry if there are none saved
+    if (!Object.keys(contentTypes).length) {
+      contentTypes[''] = { slugField: '', urlPrefix: '' };
+    }
+
+    // eslint-disable-next-line react/no-did-mount-set-state
+    this.setState(
+      {
+        allContentTypes: spaceContentTypes.reduce((acc, contentType) => {
+          contentType.fields = contentType.fields
```
jelz: I wouldn't mutate `contentType`, I'd do `{...contentType, fields: contentType.fields.filter(... }` instead

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+        <div className={styles.background} />
+        <div className={styles.body}>
+          <div className={styles.section}>
+            <Typography>
+              <Heading className={styles.spaced}>About Google Analytics</Heading>
+
+              <Paragraph>
+                This app allows you to view pageview analytics of a Contentful entry in the editor
+                sidebar. For installation instructions, please refer to the app&apos;s
+                <a href>documentation</a>.
```
davidfateh: Pretty safe bet the URL will be:

```suggestion
                <a href="https://www.contentful.com/developers/docs/extensibility/apps/google-analytics/">documentation</a>.
```

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+    );
+
+    app.onConfigure(() => this.configureApp());
+  }
+
+  async configureApp() {
+    const { contentTypes, clientId, viewId } = this.state;
+
+    const EditorInterface = Object.keys(contentTypes).reduce((acc, id) => {
+      acc[id] = { sidebar: { position: 0 } };
```
jelz: I'd do `1` so it goes below the publish button

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+                  hasError={!key}
+                  onChange={event => this.handleContentTypeChange(key, event.target.value)}>
+                  {key ? null : (
+                    <Option disabled value="">
+                      {' '}
+                      Select a Content Type{' '}
+                    </Option>
+                  )}
+
+                  {Object.entries(allContentTypes)
```
davidfateh: It would be nice to get this list in a-z order.

marketplace/google-analytics-app/src/Timeline.tsx
```diff
@@ -0,0 +1,132 @@
...
+    return !this.state.timeline || nextRange.start !== start || nextRange.end !== end;
+  }
+
+  async componentDidMount() {
+    let viewUrl = '';
+
+    try {
+      const accounts = (await gapi.client.analytics.management.accountSummaries.list()) || [];
+      viewUrl = this.getExternalUrl(accounts);
+      // console.log(accounts)
```
jelz: remove

marketplace/google-analytics-app/src/Timeline.tsx
```diff
@@ -0,0 +1,132 @@
...
+    } catch (error) {
+      console.log(error);
```
jelz: Display a meaningful information to the user

marketplace/google-analytics-app/src/index.spec.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import React from 'react';
+import { render, fireEvent, cleanup, configure } from '@testing-library/react';
+import { AppConfig, SidebarExtension } from './index';
+
+configure({
+  testIdAttribute: 'data-test-id'
+});
+
+describe('DialogExtension', () => {
```
jelz: Remove this file

marketplace/google-analytics-app/src/Timeline.tsx
```diff
@@ -0,0 +1,132 @@
+/* global gapi */
```
davidfateh: We should instead pass `gapi` into the component through props. This way we can eliminate more global variables.

marketplace/google-analytics-app/src/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  componentDidMount() {
+    this.props.sdk.window.startAutoResizer();
+    const { auth } = window.gapi.analytics;
+
+    auth.on('signIn', () => this.setState({ isAuthorized: true }));
+    auth.on('signOut', () => this.setState({ isAuthorized: false }));
+  }
+
+  onButtonClick = () => {
```
jelz: This should be gone

marketplace/google-analytics-app/src/Timeline.tsx
davidfateh: Remove

marketplace/google-analytics-app/src/Timeline.tsx
davidfateh: We should probably have a better UI for this with either showing an error message in the sidebar or using our notifier.

```suggestion
  this.props.sdk.notifier.error('Failed to load your Google Analytics data');
```

marketplace/google-analytics-app/src/Timeline.tsx
```diff
@@ -0,0 +1,132 @@
...
+          <SkeletonContainer>
+            <SkeletonImage width={window.innerWidth} height={CHART_HEIGHT} />
+          </SkeletonContainer>
+        )}
+      </div>
+    );
+  }
+}
+
+Timeline.propTypes = {
```
davidfateh: I would suggest using TS in this component since we use it elsewhere. Seems strange to have a mix in this package.

marketplace/google-analytics-app/src/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+    return this.props.sdk.dialogs.openExtension({
+      width: 800,
+      title: 'The same extension rendered in modal window'
```
davidfateh: This seems like a strange title.

marketplace/google-analytics-app/src/index.tsx
```diff
@@ -0,0 +1,103 @@
...
+    if (!isAuthorized) {
+      const renderAuthButton = authButton => {
+        window.gapi.analytics.auth.authorize({
+          container: authButton,
+          clientid: this.props.sdk.parameters.installation.clientId
+        });
+      };
+
+      return (
+        <div ref={renderAuthButton} clsas={isAuthorized ? styles.hidden : styles.signInButton} />
```
davidfateh:
```suggestion
        <div ref={renderAuthButton} className={isAuthorized ? styles.hidden : styles.signInButton} />
```

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,302 @@
...
+      if (key && saved) {
+        acc[key] = saved;
+      }
+
+      return acc;
+    }, {});
+
+    // add an incomplete contentType entry if there are none saved
+    if (!Object.keys(contentTypes).length) {
+      contentTypes[''] = { slugField: '', urlPrefix: '' };
```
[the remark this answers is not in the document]
taye: yes

marketplace/google-analytics-app/src/AppConfig.tsx
taye: This is only to avoid the order of the rows of content types changing because of a change in the output `Object.keys()` when the user selects a different content type for a row. I could this properly by maintaining an array of selected keys, but this is much easier and only breaks with all-number keys.

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,342 @@
...
+                  isDisabled={!key}
+                  hasError={key ? !slugField : false}
+                  value={slugField}
+                  onChange={(event: React.ChangeEvent<HTMLInputElement>) =>
+                    this.handleContentTypeFieldChange(key, 'slugField', event.target.value)
+                  }>
+                  <Option {...optionPlaceholderAttrs} value="">
+                    Select slug field
+                  </Option>
+                  {key
```
jelz: easier for eyes: `{key && allContentTypes[key].fields.map(...`

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,342 @@
...
+      delete contentTypes[key];
+
+      return { contentTypes };
+    });
+  }
+
+  render() {
+    const { contentTypes, allContentTypes } = this.state;
+
+    const optionPlaceholderAttrs: { [key: string]: boolean } = { disabled: true, selected: true };
```
jelz: This looks like something that could live outside of the component. And the typing I think is not needed, should be inferred.

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,342 @@
...
+    if (!Object.keys(contentTypes).length) {
+      contentTypes[''] = { slugField: '', urlPrefix: '' };
+    }
+
+    // eslint-disable-next-line react/no-did-mount-set-state
+    this.setState(
+      {
+        // sort contentTypes by display name
+        allContentTypes: sortBy(spaceContentTypes, 'name').reduce(
+          (acc: AppConfigState['allContentTypes'], contentType) => {
```
jelz: I'd introduce a separate type instead of using `AppConfigState['allContentTypes']`

marketplace/google-analytics-app/src/AppConfig.tsx
```diff
@@ -0,0 +1,342 @@
...
+
+  handleContentTypeChange(prevKey: string, newKey: string) {
+    this.setState(prevState => {
+      const contentTypes: AppConfigState['contentTypes'] = {};
+
+      // remove contentType[prevKey] field and replace with the new contentType
+      // key while preserving key order
+      for (const [prop, value] of Object.entries(prevState.contentTypes)) {
+        if (prop === prevKey) {
+          contentTypes[newKey as keyof typeof contentTypes] = {
```
jelz: won't it simply accept `newKey` as a string?

marketplace/google-analytics-app/src/Timeline.tsx
```diff
@@ -0,0 +1,164 @@
...
+    const { sdk, gapi } = this.props;
+
+    try {
+      const accounts = (await gapi.client.analytics.management.accountSummaries.list()) || [];
+      viewUrl = this.getExternalUrl(accounts);
+    } catch (e) {
+      const error = e.result ? e.result.error : e;
+      sdk.notifier.error(
+        `Google Analytics App couldn't get a link to your dashboard (${error.message})`
+      );
```
jelz: Should we continue here or `return` early?

marketplace/google-analytics-app/src/Timeline.tsx
```diff
@@ -0,0 +1,164 @@
...
+    timeline.on('success', this.onSuccess);
+    timeline.on('error', this.onError);
+
+    // eslint-disable-next-line
+    this.setState({ timeline, viewUrl });
+    this.updateTimeline();
+  }
+
+  componentDidUpdate(prevProps: TimelineProps) {
+    for (const key of ['dimensions', 'start', 'end', 'pagePath'] as const) {
```
jelz: the array here could be extracted and given a meaningful name

marketplace/google-analytics-app/src/Timeline.tsx
```diff
@@ -0,0 +1,164 @@
...
+      if (this.props[key] !== prevProps[key]) {
+        this.updateTimeline();
+        break;
+      }
+    }
+  }
+
+  updateTimeline() {
+    if (!this.state.timeline) {
```
jelz: I wouldn't set `timeline` on the state, `this.timeline` should be fine.

marketplace/google-analytics-app/src/Timeline.tsx
```diff
@@ -0,0 +1,164 @@
...
+      'start-date': formatDate(start),
+      'end-date': formatDate(end)
+    };
+
+    this.state.timeline.set({ query }).execute();
+    this.setState({ loading: true });
+    this.props.onQuery();
+  }
+
+  getExternalUrl(accounts: {
```
jelz: introduce a type for this thing

marketplace/google-analytics-app/src/Timeline.tsx
taye: The try block is only to get the url to the report in the google analytics website, so it's fine to continue without it. If we can't get the url, then we just don't render a link.

marketplace/google-analytics-app/src/AppConfig.tsx
taye: No, because `contentTypes` has no `[key: string]: SpecificType` decleration, so `contentTypes[string]` is `any`.

marketplace/google-analytics-app/src/AppConfig.tsx
taye: The problem here was that the `<Options>` component from forma doesn't have props for those attributes on the HTMLOptionElement. It turns out regular `<option>` elements work as children to forma's `<Select>` component so I've now used those.

PR #280, opened 2020-01-31
Add simple branch support for Netlify builds

swey:
It it's circumstantial and unnecessary to create different Netlify sites to build and deploy different branches (e.g. staging and production).

This PR adds an input field for the branch (no auto completion) and passes this branch to the build hook creation.

Please note that this change is untested because I have no access to the early access programm yet. But in any case it can be an inspiration for extending the existing plugin.

### DIFF
--- a/marketplace/netlify-build/src/app/netlify-client.js
+++ b/marketplace/netlify-build/src/app/netlify-client.js
@@ -120,9 +120,10 @@ export async function listSites(accessToken) {
   };
 }
 
-export function createBuildHook(siteId, accessToken) {
+export function createBuildHook(siteId, accessToken, branch = null) {
   return post(`/sites/${siteId}/build_hooks`, accessToken, {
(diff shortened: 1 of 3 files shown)


marketplace/netlify-build/src/app/netlify-client.js
```diff
@@ -120,9 +120,10 @@ export async function listSites(accessToken) {
   };
 }
 
-export function createBuildHook(siteId, accessToken) {
+export function createBuildHook(siteId, accessToken, branch = null) {
   return post(`/sites/${siteId}/build_hooks`, accessToken, {
-    title: 'Contentful integration'
+    title: 'Contentful integration' + branch ? ` (${branch})` : '',
+    branch
```
swey: cf. https://docs.netlify.com/configure-builds/build-hooks/#parameters

PR #287, opened 2020-02-27
removing extensions that were replaced by app functionality

davidfateh:
Removing extensions that have been replaced by app functionality.


suevalov:
−224,132 nice contribution :) 

are all deployed marketplace extensions/apps staying live?

hwartig:
@suevalov and ~195500 lines of the 224000 removed lines came from package-lock.json files 😳

davidfateh:
> are all deployed marketplace extensions/apps staying live?

My guess is yes since I doubt there is a process to "unbuild" them automatically. I would also assume they will remain live at the last version they were deployed at.

 I have kept a few apps which I know are potentially not migrated yet or are being used by 3rd parties.

kdamball:
Are all extensions removed, not being used by 3rd parties?

